PR #2, opened 2016-03-08
Add Subscription Tracking API spec

maxyengine:


adamaltman:
The paths should use `@` as a separator.

spec/paths/tracking@subscriptions.yaml
```diff
@@ -0,0 +1,17 @@
+get:
+  tags:
+    - Subscription
+    - Logs
+  summary: Retrieve a list of tracking subscription logs
+  parameters:
```
adamaltman: no `filter`?  I imagine a common use case would be to filter by subscription id.

spec/paths/tracking@subscriptions.yaml
```diff
@@ -0,0 +1,17 @@
+get:
+  tags:
+    - Subscription
+    - Logs
```
slavcodev: Maybe rename tag to `Tracking`?

spec/definitions/SubscriptionLog.yaml
```diff
@@ -0,0 +1,30 @@
...
+    description: Subscription's result
+    type: string
+    enum:
+      - created
+      - postponed
+      - stopped
+      - error
+    readOnly: true
+  message:
+    type: string
```
IvanGoncharov: @slavcodev Short answer no.
It's purely Swagger extension to JSON Schema. So it fully described here:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-13
> Relevant only for Schema "properties" definitions.

slavcodev: OK. Thanks.

maxyengine: yes you're right

spec/paths/tracking@subscriptions.yaml
maxyengine: Ok. So and for API tracking as well

spec/paths/tracking@subscriptions@{id}.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - $ref: "#/parameters/resourceId"
+get:
+  tags:
+    - Tracking
+  summary: Retrieve a tracking subscription log with specified identifier string
+  responses:
+    200:
+      description: Tracking subscription log was retrieved successfully
+      schema:
+        $ref: "#/definitions/SubscriptionLog"
```
adamaltman: if the endpoint is renamed to tracking/subscriptions perhaps the api definition should be SubscriptionTracking instead of SubscriptionLog.

maxyengine: Then i'll rename ApiLog to ApiTracking as well in this PR

spec/definitions/SubscriptionTracking.yaml
```diff
@@ -0,0 +1,32 @@
+type: object
+description: Tracking subscription log
+properties:
+  id:
+    $ref: "#/definitions/ResourceId"
```
slavcodev: That's wrong. Specification permits only `$ref`, another attributes ignore. There is a "allOf" case, look examples in existing definitions.

spec/definitions/SubscriptionTracking.yaml
```diff
@@ -0,0 +1,32 @@
...
+    readOnly: true
```
slavcodev: Seems all `SubscriptionTracking` object is `readOnly`

spec/paths/tracking@subscriptions.yaml
```diff
@@ -0,0 +1,16 @@
+get:
+  tags:
+    - Tracking
+  summary: Retrieve a list of tracking subscription logs
+  parameters:
+    - $ref: "#/parameters/collectionLimit"
+    - $ref: "#/parameters/collectionOffset"
+  responses:
+    200:
```
slavcodev: Missing `rate limiting` and `pagination` headers

slavcodev:
@maxyengine Let's group the tracking definitions in the folder `Tracking`, like [`Criteria`](https://github.com/Rebilly/RebillyAPI/tree/master/spec/definitions)

PR #3, opened 2016-03-08
Added pathes and definition for api logs

maxyengine:


slavcodev:
In this repository, was changed file name convention for `paths`  definitions. We should use `@` instead of `.` (dot)

spec/definitions/ApiLog.yaml
```diff
@@ -0,0 +1,48 @@
...
+    type: string
+    enum:
+      - HEAD
+      - GET
+      - POST
+      - PUT
+      - DELETE
+    readOnly: true
+  ipAddress:
+    description: Request IP address
```
slavcodev: It's `Client` IP address?

maxyengine: yes

slavcodev:
You're using tag `Tracking`, but it's not exists. Please also add it definition.

slavcodev:
:+1:

PR #5, opened 2016-03-12
Massive initial edit...

adamaltman:
- added more detailed descriptions for some tags
- removed use of Public tag
- changed tags to plural (natural language)
- added more collection parameters, and implemented for the get customers resource


slavcodev:
Looks great. Just would like to suggest to avoid double spaces :)

adamaltman:
Good habits die hard... :wink: 

This is really a work in progress, but since the spec is actively being worked on, I thought it would be better to merge this sooner.  The remainder will be smaller edits.

PR #7, opened 2016-03-16
Payment method discriminator

colinbird:


RomanGotsiy:
Unfortunately, this schema is incorrect.

That's because how swagger discriminator works (from my point of view it works stupid)

from the Swagger spec:
> When used, the discriminator will be the name of the property used to decide which schema definition is used to validate the structure of the model ... The **value** of the chosen property has to be the **friendly name given to the model under the definitions property**"

What does it mean? It means that schema you prepared in this PR supposes that values of `payment` fields should be: `PaymentAch`, `PaymentCardPayment`. But this is wrong because actual values of `payment` field are `payment_card` and `ach`.

So to describe this you should name your "derived" definitions as `payment_card` and `ach` which I agree is not very smooth but that's the only way.

PR #9, opened 2016-03-21
Add Webhook Tracking API spec

maxyengine:


slavcodev:
@adamaltman there is question, we've discussed to implement new Webhook API soon, I think some parts of this tracking will be changed, e.q. `websiteName` seems will disappear. what do you think? 
Should we release this API as is or should wait?
If we should release this ASAP, may be at least use another endpoint, to reserve `/tracking/webhooks/` to new API?

adamaltman:
I'm all for that.  In fact, we can wait for the new API... as long as we think the new API spec will be within the next 4-6 weeks.

spec/paths/tracking@webhooks.yaml
```diff
@@ -0,0 +1,29 @@
+get:
+  tags:
+    - Tracking
+  summary: Retrieve a list of tracking webhook notifications
+  parameters:
+    - $ref: "#/parameters/collectionLimit"
+    - $ref: "#/parameters/collectionOffset"
+  responses:
+    200:
+      description: Tracking webhook notifivations was retrieved successfully
```
creocoder: Typo there, notifivations => notifications

PR #6, opened 2016-03-21
Members other than "$ref" in a JSON Reference

slavcodev:
In many places like [this](https://github.com/Rebilly/RebillyAPI/blob/master/spec/definitions/Customer.yaml#L6):

```yaml
id:
  description: The customer identifier string
  $ref: "#/definitions/ResourceId"
  readOnly: true
```

the same mistake is present: other members (in this particular case `description` and `readOnly`) are used on the same level as `$ref`. But according to the [JSON Reference RFC](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-02#section-3) these fields shall be ignored:

> Any members other than "$ref" in a JSON Reference object SHALL be ignored.

To fix this you can use `allOf`.

```yaml
id:
  description: The customer identifier string
  readOnly: true
  allOf:
```

----
In theory the following way is also correct:

```yaml
id:
  allOf:
    - description: The customer identifier string
      readOnly: true
```

but in this case it is not defined which description is more prioritized


slavcodev:
@RomanGotsiy please review this one.

spec/definitions/ApiLog.yaml
```diff
@@ -41,8 +42,7 @@ properties:
     readOnly: true
   createdTime:
     description: The log created time
-    $ref: "#/definitions/ServerTimestamp"
-    readOnly: true
```
RomanGotsiy: I believe you accidentally removed `readOnly: true`

RomanGotsiy:
@slavcodev other changes seem good :+1:

spec/definitions/ApiLog.yaml
slavcodev: Nope. `ServerTimestamp` define all object as read only.

PR #15, opened 2016-03-23
Api Keys Management

maxyengine:


spec/definitions/ApiKey.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - $ref: "#/definitions/ResourceId"
+  description:
+    description: API key description
+    type: string
+  datetimeFormat:
+    description: Date time format
+    type: string
+    enum:
+      - mysql
+      - iso8601
```
adamaltman: should default to iso8601

spec/definitions/ApiKey.yaml
```diff
@@ -0,0 +1,30 @@
...
+properties:
+  id:
+    readOnly: true
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  description:
+    description: API key description
+    type: string
+  datetimeFormat:
+    description: Date time format (default to iso8601)
```
slavcodev: Swagger supports `default` property for model

maxyengine: ah Ok thanks

PR #20, opened 2016-04-05
Checkout Page spec

XaosSintez:


spec/definitions/CheckoutPage.yaml
```diff
@@ -0,0 +1,42 @@
...
+  isActive:
+    description: If checkout page active
+    type: boolean
+  redirectUrl:
+    description: Checkout page url
+    type: string
+  redirectTimeout:
+    description: Checkout page redirect timeout
+    type: integer
+  customCustomerId:
```
slavcodev: looks weird the property name for boolean type.

XaosSintez: how about `allowCustomCustomerId` ?

spec/definitions/CheckoutPage.yaml
```diff
@@ -0,0 +1,42 @@
...
+    description: Checkout page plan ID
+    type: string
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  websiteId:
+    description: Checkout page website ID
+    type: string
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  uri:
```
slavcodev: Actually there is not the `URI`, maybe another name? `uriPath`? `slug` (I'm not sure this name is clear)?

XaosSintez: Changed to `uriPath` and added better description

PR #24, opened 2016-04-11
Rules API

slavcodev:
You can see docs preview [here](http://rebilly.github.io/RebillyAPI/preview/rules/).


slavcodev:
screenshot

spec/definitions/Rules/RuleAction.yaml
```diff
@@ -0,0 +1,27 @@
...
+      - cancel-scheduled-payments
+      - close-invoices
+      - dun-payment
+      - guess-payment-card-expiration
+      - pick-gateway-account
+      - schedule-dunning
+      - schedule-payment
+      - stop-subscriptions
+  active:
+    description: Whether rule is active
```
adamaltman: Whether `action` is active, right?

slavcodev: yes

creocoder:
I think its OK. Not sure strict is necessary.

slavcodev:
Updated conditions

Screenshot

slavcodev:

Teknologica:
For reference, here's a criteria structure for `amount === 50 || (amount !== 100 && amount > 60)`
```js
criteria: {
    operation: 'or',
    conditions: [
        {
            operation: 'and',
            conditions: [
                {
                    operation: 'not',
                    condition: {
                        operation: 'equals',
                        field: 'amount',
                        value: '100'
                    }
                },
                {
                    operation: 'gt',
                    field: 'amount',
                    value: '60'
                }
            ]
        },
        {
            operation: 'equals',
            field: 'amount',
            value: '50'
        }
    ]
}
```

spec/definitions/Rules/Actions/schedule-payment.yaml
```diff
@@ -0,0 +1,27 @@
...
+          - immediately
+          - after-time
+          - specific-time
+      scheduleAfterTime:
+        type: integer
+      scheduleSpecificTime:
+        type: string
+        enum:
+          - invoice-due-time
+      # TODO: Consider to change name, e.q. amountInstruction, amountPolicy
```
XaosSintez: I think `amountPolicy` is good

spec/definitions/Rules/EmailNotification.yaml
```diff
@@ -0,0 +1,41 @@
...
+      format: email
+  subject:
+    type: string
+  bodyText:
+    type: string
+    description: Leave empty to use text from "bodyHtml" without tags
+  bodyHtml:
+    type: string
+    description: Leave empty to recieve "text/plain" email
+required:
```
XaosSintez: We usually put `required` section on the top as i remember

spec/definitions/Rules/Rule.yaml
```diff
@@ -0,0 +1,21 @@
+type: object
+description: The rule
+properties:
+  name:
+    type: string
+  status:
+    $ref: "#/definitions/OnOff"
+  final:
```
XaosSintez: may be `isFinal` ? Not sure if it's better

spec/paths/rules.yaml
```diff
@@ -0,0 +1,18 @@
+get:
+  tags:
+    - Rules
+  summary: Retrieve a list of rules sets
+  description: |
+    Retrieve a list of rules sets
+  parameters:
+    - $ref: "#/parameters/collectionLimit"
+    - $ref: "#/parameters/collectionOffset"
+  responses:
```
XaosSintez: You can add here headers

spec/definitions/Rules/EmailNotification.yaml
slavcodev: I didn't know we have a convention here. So, It's my personal preference, to read which properties exist, and then which are required. I can change it, if it necessary.

spec/definitions/Rules/Rule.yaml
slavcodev: `isFinal` looks like answer, like action to get this value but `final` looks like property name. That was my though.

spec/paths/rules.yaml
slavcodev: Sure 👍

spec/definitions/Rules/Rule.yaml
Teknologica: `final` is good, otherwise might need something more semantic like `stopsExecution`, `stopChain` or `stopSequence`

spec/definitions/Rules/Actions/schedule-payment.yaml
```diff
@@ -0,0 +1,27 @@
+description: ""
+allOf:
+  - $ref: "#/definitions/RuleAction"
+  - type: object
+    properties:
+      # TODO: Consider to change name, e.q. scheduleInstruction, schedulePolicy
```
Teknologica: If not `schedule` I personally like `schedulePolicy`

adamaltman: `trigger-payment`

spec/definitions/Rules/EmailNotification.yaml
```diff
@@ -0,0 +1,41 @@
+type: object
+properties:
+  title:
```
adamaltman: Should the `title` be `name`, or in Webhooks should `name` be `title`?

spec/definitions/Rules/Webhook.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - PATCH
+      - DELETE
+  uri:
+    type: string
+    format: uri
+  query:
+    type: object
+    description: The URI parameters
+    additionalProperties:
+      type:
```
RomanGotsiy: Unfortunately `type` [should be a scalar](https://github.com/OAI/OpenAPI-Specification/issues/333#issuecomment-206420760) in Swagger 2.0. If it is possible to somehow omit array `type` do it please, otherwise we can discuss it.

@IvanGoncharov has already pushed a proposal to fix this in the next Swagger version:

spec/definitions/Rules/WebhookAuthorization/oauth2.yaml
```diff
@@ -0,0 +1,3 @@
+description: ""
+allOf:
```
adamaltman: actually, I think these authorization types are missing the proper definitions.

spec/paths/rules@{event}.yaml
```diff
@@ -0,0 +1,47 @@
...
+    Update the rules for specified event
+  parameters:
+    - name: body
+      in: body
+      description: Set of rules resource
+      required: true
+      schema:
+        $ref: "#/definitions/RuleSet"
+  responses:
+    200:
```
dara123: Should we also have 201 when they don't have any rules set yet?

adamaltman: The event exists, so they always have a ruleset.  Like an empty collection always returns a 200.

spec/definitions/Rules/Webhook.yaml
adamaltman: please +1 that issue...

spec/definitions/Rules/Actions/pick-gateway-account.yaml
```diff
@@ -0,0 +1,22 @@
+description: ""
+allOf:
+  - $ref: "#/definitions/RuleAction"
+  - type: object
+    properties:
+      listType:
+        type: string
+        enum:
+          - gateway-account
```
adamaltman: Throughout the api we use different naming conventions: `snake_case` and `kebab-case` and `camelCase`.  It's not just here, throughout the actions you use `kebab-case` for field/member values, and `camelCase` for field/member names.  Fortunately, we're pretty consistent with field/member names.  

However, for values, we're not.  The rules api is pretty large addition to the overall api schema.  I think a decision should be made about these values, and we should go and adjust the spec to use the selected values, and make sure we support both for backwards compatibility.

spec/paths/rules@{event}.yaml
```diff
@@ -0,0 +1,47 @@
+parameters:
+  - name: event
+    in: path
+    description: The resource identifier string
+    type: string
+    required: true
```
adamaltman: hmm, shouldn't we list an enum of these?  Or we can't because of custom events?

spec/definitions/Rules/Actions/schedule-payment.yaml
slavcodev: @adamaltman what do you mean? did you suggest to rename action name? my to-do comment regarding attribute name, and the definitions of this property changed in #35

spec/definitions/Rules/EmailNotification.yaml
slavcodev: I do not know. Should it be? For me is `webhook name` and `notification title` :)

spec/definitions/Rules/Webhook.yaml
slavcodev: @creocoder what d you think? how we can change this definition to avoid multiple types?

spec/definitions/Rules/Actions/pick-gateway-account.yaml
slavcodev: Totally agreed. I'm hoping, I was consistent in `Rules API` and use

- camelCase: for attribute names
- kebab-case: for enum values

If I missed something let me know, please. Or if you having arguments against this naming, let us know, I'll change quickly.

spec/paths/rules@{event}.yaml
slavcodev: Hmm, I don't know if swagger supports enum in parameters, I think it supports only in resources schemas.

@RomanGotsiy @IvanGoncharov do you can help here?

spec/paths/rules@{event}.yaml
slavcodev: Yes. Adam is right. Here we never will create anything. The rules sets are always exists for system events.

spec/paths/rules@{event}.yaml
IvanGoncharov: @slavcodev It supported 
```yaml
parameters:
  - name: event
    in: path
   description: The resource identifier string
   required: true
   type: array
   items:
     type: string
     enum: 
        - value1
        - value2
```
But if delimiter is something different that `,` you should specify correct `collectionFormat`:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#user-content-parameterCollectionFormat

slavcodev:
Some notes which may help to understand API

> PUT https://api.rebilly.com/v2.1 /rules/{event} ... this is to create and update a set of rules, right?

Nope, this request only will update set of rules, because by design, set of rules for system (pre-defined) events always exists, and initially it empty

> I guess the "version" attribute is an incrementing value that starts at 0? And any changes to the rule set's Rule attributes (criteria, final, status, actions, action status, action name, etc.) will increment that version number?

That's right. Which system event has set of rules. Initial version of rules equals to zero. Any changes for this set of rules, will change the set's version. The merchant able to read all versions, history saved.

> events seem hard coded for the most part but there will be custom events too

I believe `rules of system events` and `rules of custom events` should have different endpoints. There are my thoughts

1. Mixing `rules/<eventType>` for system, and `rules/<customEventId>` seems wrong. System event types are pre-defined, you can see enum in spec, but `custom event id` is dynamic, merchant create, able to remove, so IDs will change from time to time.

2. Access control issue. Endpoints should be different, to able to configure permissions to this API.

ping: @Zaphell

spec/definitions/Rules/Actions/pick-gateway-account.yaml
Zaphell: It would be nice to have that consistency throughout the whole API, not just the Rules API (if possible). I don't know of any reason not to use ```kebab-case```, so no objections there.

spec/definitions/Rules/Webhook.yaml
creocoder: @slavcodev It should be similar to headers.

slavcodev: @creocoder It cannot be similar to headers. Headers also should be change, because multiple types don't supported.

spec/definitions/Rules/WebhookAuthorization/oauth2.yaml
slavcodev: Yes, it's placeholders.

@creocoder Are the auth types having final specs?

creocoder: @slavcodev Yes, will write to HitChat to discuss.

adamaltman:
Can we merge this?

slavcodev:
New issues:

- [ ] Rename "operation" to "op" in criteria
- [ ] Review criteria, to allow deep level conditions
- [ ] Rename "field" to "path" in criteria, and use JSON pointer

spec/definitions/Rules/RuleSet.yaml
```diff
@@ -0,0 +1,22 @@
...
+    description: Event description
+    readOnly: true
+  version:
+    type: integer
+    readOnly: true
+  rules:
+    type: array
+    items:
+      $ref: "#/definitions/Rule"
+  updatedTime:
```
adamaltman: seems it would be good to have `updatedBy` here too.

slavcodev: yes, maybe. I deliberately did not add this and `createdBy` attributes in start API, which we've discussed in `Rule RFC` doc. We may add this later, w/o to brake compatibility. But for now I d not know what to store here? User Id based on auth (API Key)? Or what?
I think would be more useful to have audit log, who and what resource changed, instead of these attributes. I really do not understand purpose of it.

spec/definitions/Rules/ScheduledDunning.yaml
```diff
@@ -0,0 +1,48 @@
...
+          $ref: "#/definitions/ScheduleInstruction"
+        paymentInstruction:
+          type: object
+          properties:
+            amountPolicy:
+              type: string
+              enum:
+                - "partial-amount"
+                - "rate-discount"
+                # TODO: Unsupported at the moment
```
adamaltman: what about the full/normal price retry?

spec/definitions/Rules/EmailNotification.yaml
```diff
@@ -0,0 +1,42 @@
+type: object
+properties:
+  name:
+    description: The notification name
+    type: string
+  status:
+    $ref: "#/definitions/OnOff"
+  credentialId:
```
vova07: As I remember we decide to not allow merchants to send emails via `Rebilly` credentials.
I think it must be renamed to `sender` instead.

rebillyPM:
Not sure if it is too late, but the naming could be improved for consistency and clarity.  For example, instead of "amount" of days and hours: 
- number-of-days.yaml
- number-of-hours.yaml

And, weekday in English refers to the typical workdays of the week, Monday - Friday, so specifying weekday.yaml to include all seven days of the week is confusing.  To be consistent with "day-of-month.yaml" this would be more clear:
- day-of-week.yaml
 

screenshot_061316_040954_pm

slavcodev:
Thanks @rebillyPM for great review. I believe it is not late and we can change it.

slavcodev:
~~~
format: time
~~~

[is invalid type](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types)

spec/definitions/Rules/SmtpCredentials.yaml
```diff
@@ -0,0 +1,38 @@
...
+  authenticationMethod:
+    description: The authentication emthod
+    type: string
+    default: none
+    enum:
+      - none
+      - plain
+      - logic
+      - cram-md5
+required:
```
slavcodev: It's seems wrong, for `none` which is default, all this attributes not used.

spec/definitions/Rules/SmtpCredentials.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - tls
+      - ssl
+  authenticationMethod:
+    description: The authentication emthod
+    type: string
+    default: none
+    enum:
+      - none
+      - plain
+      - logic
```
slavcodev: typo `login`

spec/definitions/Rules/SmtpCredentials.yaml
```diff
@@ -22,17 +30,9 @@ properties:
...
-    enum:
-      - none
-      - plain
-      - logic
-      - cram-md5
 required:
   - username
   - password
   - host
+  - authenticationMethod
```
slavcodev: Removing default value and making this attribute do not resolve the issue, the `username` and `password` make no sense for `none` auth method, so they can not be required.

spec/definitions/PaymentRetry/PaymentRetry.yaml
```diff
@@ -19,13 +19,13 @@ properties:
...
+    description: The policy on the payment retry ends
     type: string
     enum:
       - "none"
       - "cancel-subscription"
 required:
   - attempts
   - afterAttemptPolicy
-  - afterDunningEndPolicy
+  - afterPaymentRetryEndPolicy
```
slavcodev: In this case, I think we can omit the `payment`, so `afterRetryEndPolicy` seems nice enough. Thoughts?

spec/definitions/Rules/UserPasswordAuthorization.yaml
```diff
@@ -0,0 +1,10 @@
+type: object
+properties:
+  username:
+    type: string
+  password:
```
adamaltman: This would change to `passwordToken`.  And the string it would accept would be the token of the `passwords` resource.

spec/definitions/Rules/WebhookAuthorization/basic.yaml
```diff
@@ -0,0 +1,12 @@
+allOf:
+  - $ref: "#/definitions/WebhookAuthorization"
+  - type: object
+    properties:
+      username:
+        type: string
+      password:
```
adamaltman: Another `passwordToken`

spec/definitions/Rules/WebhookAuthorization/digest.yaml
```diff
@@ -0,0 +1,12 @@
+allOf:
+  - $ref: "#/definitions/WebhookAuthorization"
+  - type: object
+    properties:
+      username:
+        type: string
+      password:
```
adamaltman: Another `passwordToken`

spec/definitions/Rules/WebhookAuthorization/oauth1.yaml
```diff
@@ -0,0 +1,17 @@
+allOf:
+  - $ref: "#/definitions/WebhookAuthorization"
+  - type: object
+    properties:
+      consumerKey:
+        type: string
+      consumerSecret:
```
adamaltman: I think this should probably be another `passwordToken` type field? as well as `tokenSecret`?

creocoder: Its better to use standartized oauth 1 terms there.

spec/definitions/Rules/UserPasswordAuthorization.yaml
vova07: How about just `token` instead of `passwordToken` ?

adamaltman:
``` Error: WebhookAuthorization definition already exists
    at /home/travis/build/Rebilly/RebillyAPI/node_modules/swagger-repo/lib/index.js:232:13
    at arrayReduce (/home/travis/build/Rebilly/RebillyAPI/node_modules/swagger-repo/node_modules/lodash/lodash.js:687:21)
    at Function.reduce (/home/travis/build/Rebilly/RebillyAPI/node_modules/swagger-repo/node_modules/lodash/lodash.js:9545:14)
    at globObject (/home/travis/build/Rebilly/RebillyAPI/node_modules/swagger-repo/lib/index.js:229:12)
    at globYamlObject (/home/travis/build/Rebilly/RebillyAPI/node_modules/swagger-repo/lib/index.js:240:22)
    at Object.exports.bundle (/home/travis/build/Rebilly/RebillyAPI/node_modules/swagger-repo/lib/index.js:127:27)
    at Command.<anonymous> (/home/travis/build/Rebilly/RebillyAPI/node_modules/swagger-repo/bin/swagger-repo.js:41:23)
    at Command.listener (/home/travis/build/Rebilly/RebillyAPI/node_modules/commander/index.js:301:8)
    at emitTwo (events.js:106:13)
    at Command.emit (events.js:191:7)
npm ERR! Test failed.  See above for more details.
```

slavcodev:
Congrats it's merged. Thanks @all for contributing! ✌️

PR #25, opened 2016-04-11
Rules alternative version of conditions

slavcodev:
[Alternative version of conditions](https://github.com/Rebilly/RebillyAPI/pull/24): combined 4 definitions in one with additional property


slavcodev:
screenshot

XaosSintez:
If we are going to remove `lt`, `gt`, `ne` then :+1:

slavcodev:
Yes, we do, `lt`, `gt` removed in this PR. But I'm ensure, is it not confused the attributes `operation` and `operator` in one object? e.q.

```json
{
   "criteria": {
     "operation": "compare",
     "operator": "=",
     "field": "customerId",
     "value": "abc",    
   }
}
```

XaosSintez:
i think it is ok. But someone can disagree with me

creocoder:
I think its OK. Not sure strict is necessary.

slavcodev:
Well, as far as I see, PR declined.

PR #28, opened 2016-04-11
Rules webhook action

slavcodev:
blocks #24

### DIFF
--- a/spec/definitions/Rules/Actions/Webhook.yaml
+++ b/spec/definitions/Rules/Actions/Webhook.yaml
@@ -1,0 +1,49 @@
+type: object
+properties:
+  name:
+    description: The webhook name
+    type: string
+  status:
+    type: string
+    default: active
+    enum:
+      - active
+      - inactive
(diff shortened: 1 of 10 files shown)


creocoder:
Ideal payload as i see that:

```json
{
    "name": "Test webhook",
    "method": "POST",
    "url": "http://example.com",
    "query": {
        "param1": "value1",
        "param2": ["value2", "value3"],
    },
       "auth":  {
           "type": "foo",
           // custom fields there depends on type
       }
    "headers": {
        "name1": "value1",
        "name2": ["value2", "value3"],
    },
    "body": "some",
    "status": "active",
}
```

Only "name", "method" and "url" is required. Other fields are optional.

slavcodev:
Screenshot

slavcodev:

adamaltman:
seems good to me :+1:

slavcodev:
OK, I'm merging it at now, if any issues will appear, will fix separately.

PR #29, opened 2016-04-15
Alternative version of negate conditions

slavcodev:
[Alternative version of negate conditions](https://github.com/Rebilly/RebillyAPI/pull/24): remove definitions duplicates for negation and add logical "not" definition


slavcodev:
Screenshot

mdomba:
would it make sense to put "not" as first operator?

slavcodev:
first where? in the operations lists in UI? I'm not sure how they sorted :)
@RomanGotsiy said me in `ReDoc` they are not sorting, in enum I wrote: `and`, `or`, `not`, but for some reason order is unclear :)

adamaltman:
It's definitely alpha sort.  I was also going to say why isn't or next to and...

adamaltman:
Can you use all operations with not?  So..

```
not => [
    and => [
        in => ['a', 'b', 'c'],
        less-than => 50,
    ]
]
```

slavcodev:
> Can you use all operations with not?

Yes, any composition

RomanGotsiy:
@slavcodev 
> @RomanGotsiy said me, that in ReDoc they are not sorting, in enum I wrote: and, or, not, but for some reason order is unclear seems alphabetical :)

I've checked once more and they are sorted, this is not intentionally. I'll fix this.

slavcodev:
Accepted.

PR #30, opened 2016-04-15
Group files to folders by context

slavcodev:
The main goal do not stir `discriminator` specifications with entities.


slavcodev:

adamaltman:
Bank Account should be in Payments.

Blacklist should probably not be in Customers.  I don't know where it should be... maybe in the root context.

Organization and Website are not part of the Customer context.  It's related to the Merchant.

The `User` context seems to represent two concerns:

* the merchant's employees (aka user of Rebilly)
* the merchant's customers users (aka customers of merchants)

There may be some need to rename these contexts.  Perhaps the one related to merchant's customers can move to the `Customer` context?

slavcodev:
> BankAccount should be in Payments (not in Customer).

Ahh, seems I confused. I though `BankClient` is a customer data, like 'contact`, `address`, `organization`

> Blacklist should probably not be in Customers. I don't know where it should be... maybe in the root context.

Yes, agreed. Blacklist may be in root. I also though it related to customer data, we are blacklist the customer by id, email, customer card, and customer fingerprint.

> Organization and Website should not be part of the Customer context. It's related to the Merchant.

Organization - see comment for Account, same reason why initially I put to customer context. 

Website for me is like a group of customers, so also may be in customer context. But perhaps I really wrong. Maybe need to add `Merchants` context and move it there. I think `Websites` context is extra.

> The User context seems to represent two concerns...
> Perhaps the one related to merchant's customers can move to the Customer context?

Agreed, missed that. I'lll change it.

adamaltman:
I think this needs to be updated.  I think the organization would still be good.  Thoughts @RomanGotsiy ?

slavcodev:
Closed as very old.

PR #33, opened 2016-04-21
Rate limiting headers

XaosSintez:


RomanGotsiy:
@XaosSintez, swagger don't have shared headers. You can reuse only whole responses.
You can't add `headers` key at the top level of swagger:

```
{
    "code": "OBJECT_ADDITIONAL_PROPERTIES",
    "message": "Additional properties not allowed: headers",
    "path": [] // <- empty path here means top level
  }```

XaosSintez:
@RomanGotsiy yes, i already gor it :)
So, the only way is to add full headers description to each collection response?

RomanGotsiy:
Yes

slavcodev:
@XaosSintez You can add common response definition, and use `allOf` in paths.

slavcodev:
@XaosSintez do you have in mind to add pagination headers in separate PR or you're missed that issue?

XaosSintez:
@slavcodev yes, i remember, going to add them in another PR. This one is more about structure, how to make headers.

If you think this PR is good, let's merge it

slavcodev:
OK, I was confuse by status in JIRA. Merged!

PR #35, opened 2016-04-21
Rules: schedule dunning and schedule payment actions

slavcodev:


slavcodev:
Screenshot

slavcodev:

slavcodev:
**Schedule payment**

screen
screen
screen

spec/definitions/Rules/ScheduledPayment.yaml
```diff
@@ -0,0 +1,15 @@
...
+type: object
+properties:
+  # TODO: Consider to change name, e.q. scheduleInstruction, schedulePolicy
+  schedule:
+    $ref: "#/definitions/ScheduledPaymentInstruction"
+  # TODO: Consider to change name, e.q. amountInstruction, amountPolicy
+  amount:
+    type: string
+    enum:
+      - AIT
```
adamaltman: These string constants don't make sense.  Let's use something that someone can read and understand.  I think one is `balance-outstanding`... and the other is `invoice-total`.  

As far as the property names go, I think we should consider a better Schedule object.  I like `scheduleInstruction` but have you considered `scheduleWhen`?

slavcodev: > I think we should consider a better Schedule object

wondering, what improvements do you have in mind.

slavcodev:
I'm going to merge this. Will continue in parent PR #24

PR #37, opened 2016-04-28
Make the tag for Api Keys consistent.

adamaltman:


spec/paths/api-keys.yaml
```diff
@@ -1,6 +1,6 @@
 get:
   tags:
-    - ApiKey
+    - Api Keys
```
slavcodev: Would not the `API` more correctly?

adamaltman: Yes, it would be better.

slavcodev:
Looks fine 👍

PR #42, opened 2016-05-20
GWY-3927: Custom Events Rules API specs

vova07:
Here is the right diff: https://github.com/Rebilly/RebillyAPI/compare/rules...GWY-3927


vova07:
Please review!

spec/definitions/CustomRules/CustomRuleSet.yaml
```diff
@@ -0,0 +1,25 @@
...
+    description: Custom event description
+    readOnly: true
+  version:
+    type: integer
+    readOnly: true
+  rules:
+    type: array
+    items:
+      $ref: "#/definitions/Rule"
+  updatedTime:
```
XaosSintez: no `createdTime`, right?

spec/definitions/Rules/Actions/pick-gateway-account.yaml
```diff
@@ -0,0 +1,22 @@
+description: ""
```
XaosSintez: Some descriptions would be nice, i think

spec/definitions/Schedule/ScheduleInstruction.yaml
```diff
@@ -0,0 +1,16 @@
+type: object
+description: "The calculation instruction of scheduled time"
+discriminator: method
+properties:
+  method:
+    type: string
```
XaosSintez: may be set default to `auto`?

spec/definitions/CustomRules/CustomRuleSet.yaml
vova07: I think it should be.
I'll add it. 
Thanks!

vova07:
I think I need to create a new PR without `slava`'s commits.

vova07:
The new PR: https://github.com/Rebilly/RebillyAPI/pull/43

spec/definitions/Schedule/ScheduleInstruction.yaml
adamaltman: I don't think so -- I believe `auto` is something special.

spec/paths/custom-rules.yaml
```diff
@@ -0,0 +1,31 @@
+get:
+  tags:
+    - Custom Rules
+  summary: Retrieve a list of rules sets
```
adamaltman: doesn't this retrieve a list of only the custom rules sets?

adamaltman: or rather custom event rules sets?

adamaltman:
It is missing the tracking for custom events too?

spec/paths/custom-rules.yaml
vova07: Yes, you are right, but I thought if we are in `Custom Rules` section, it's a clear thing that we talk only about their rule sets.

vova07:
Yes, `tracking` is miss, because after discussion with Slava, we decide that this is part of another PR.
And we must decide how we'll go with `tracking`. I mean will be it a common `API URL` or we'll go with 2 separate `endpoints`.

vova07:
If is it possible, please comment and review the new PR: https://github.com/Rebilly/RebillyAPI/pull/43
It contains only the file I added.
Thanks!

PR #43, opened 2016-05-23
Add Custom Events Rules API specs

vova07:


vova07:
#42

spec/definitions/CustomRules/CustomRuleSet.yaml
```diff
@@ -0,0 +1,27 @@
...
+    items:
+      $ref: "#/definitions/Rule"
+  createdTime:
+    $ref: "#/definitions/ServerTimestamp"
+  updatedTime:
+    $ref: "#/definitions/ServerTimestamp"
+  _links:
+    $ref: "#/definitions/Links"
+required:
+  - rules
```
slavcodev: Guessing, missing the required `eventType`

spec/paths/custom-rules@{id}.yaml
```diff
@@ -0,0 +1,88 @@
...
+        $ref: "#/definitions/CustomRuleSet"
+    401:
+      $ref: "#/responses/AccessForbidden"
+    404:
+      $ref: "#/responses/NotFound"
+    422:
+      description: Invalid data was sent
+      schema:
+        $ref: "#/definitions/InvalidError"
+delete:
```
slavcodev: I think `DELETE` is not possible.

adamaltman: Yes, I would think you would `PUT` rules: `[]`?  Can you delete a custom event?

slavcodev: Yes, we can remove rules but `PUT` empty array. But we cannot delete rule set, the rule set should exist till the event exists.

Yes, we can remove the event, but only in case the rules set is empty (no rules).

vova07:

spec/definitions/CustomRules/CustomRuleSet.yaml
```diff
@@ -0,0 +1,28 @@
...
+    description: Custom event description
+    readOnly: true
+  version:
+    type: integer
+    readOnly: true
+  rules:
+    type: array
+    items:
+      $ref: "#/definitions/Rule"
+  createdTime:
```
slavcodev: I think it is not exist.

spec/definitions/CustomRules/CustomRuleSet.yaml
```diff
@@ -0,0 +1,28 @@
+type: object
+description: Set of rules for particular event
+properties:
+  id:
```
slavcodev: Seems `readOnly` and extends `ResourceId`

spec/paths/custom-rules@{id}.yaml
```diff
@@ -0,0 +1,88 @@
...
+      headers:
+        X-Rate-Limit-Limit:
+          $ref: "#/headers/X-Rate-Limit-Limit"
+        X-Rate-Limit-Remaining:
+          $ref: "#/headers/X-Rate-Limit-Remaining"
+        X-Rate-Limit-Reset:
+          $ref: "#/headers/X-Rate-Limit-Reset"
+      schema:
+        $ref: "#/definitions/CustomRuleSet"
+    201:
```
slavcodev: Response is not achieved ever

vova07: How does the rule sets created for `Custom Events` ?
Is there an event `hook` ?
I know that `rule sets` for our `rules engine` will be created by `cron` script, but I'm confused how this logic will work for `Custom Events` ?

spec/definitions/CustomRules/CustomRuleSet.yaml
vova07: Done! 
Thanks!

spec/paths/custom-rules@{id}.yaml
slavcodev: Rule set will create in the moment the event created, i.e. each event always has rule set.

vova07: Ok, got it. Thought it will work differently than in `Rules Engine`.

PR #45, opened 2016-05-25
Add Custom Events API spec

vova07:


vova07:

spec/definitions/CustomEvents/TimeOffset.yaml
```diff
@@ -0,0 +1,15 @@
+type: object
+required:
+  - value
```
slavcodev: Is `unit` required?

spec/definitions/CustomEvents/CustomEvent.yaml
```diff
@@ -0,0 +1,27 @@
...
+  description:
+    description: The custom event description
+    type: string
+  timeOffset:
+    $ref: "#/definitions/TimeOffset"
+  createdTime:
+    $ref: "#/definitions/ServerTimestamp"
+  handledTime:
+    $ref: "#/definitions/ServerTimestamp"
+  scheduledTime:
```
slavcodev: What is difference between `createdTime` and `scheduledTime` ?

spec/definitions/CustomEvents/TimeOffset.yaml
vova07: In the code it's not.
Because by default it's set to `seconds`.
But maybe for consistency it should be.

spec/definitions/CustomEvents/CustomEvent.yaml
vova07: `createdTime` is the point of time when a `customEvent` was created.
`scheduledTime` is the point of time when that custom event was scheduled, but after it will be handled it will be null.

slavcodev: It's our domain specific logic, I don't think the API should share this knowledge.

spec/definitions/CustomEvents/TimeOffset.yaml
slavcodev: I think it's safer to require it and do not use defaults, that may confuse the merchant.

adamaltman: I like that idea. :+1:

spec/definitions/CustomEvents/CustomEvent.yaml
```diff
@@ -0,0 +1,23 @@
...
+    $ref: "#/definitions/CustomEventType"
+  title:
+    description: The custom event title
+    type: string
+  description:
+    description: The custom event description
+    type: string
+  timeOffset:
+    $ref: "#/definitions/TimeOffset"
+  createdTime:
```
adamaltman: can it be updated?  seems it could?  so we should present the `updatedTime` too.

slavcodev: yes, perhaps, I removed temporary `updatedTime`. we can add later, but not on release, to not make bad decision on backend implementation.

spec/paths/custom-events@{id}.yaml
```diff
@@ -0,0 +1,98 @@
...
+    Create or update a custom event with predefined identifier string
+  parameters:
+    - name: body
+      in: body
+      description: Custom event resource
+      required: true
+      schema:
+        $ref: "#/definitions/CustomEvent"
+  responses:
+    200:
```
vova07: @adamaltman have a question about this response because it's not clear to me.
Are we able to modify a `handled` custom event ?
If yes, I think we are able to modify only `title` and `description` or not ? 
I see in the code of my use cases that we are not able to modify any `handled` events, so I need to know how will go?
Thanks!

adamaltman: I think not modifying handled events is good.

slavcodev: What is the reason to deny editing of title and description? I think, merchant might make typo, and will want to fix it when event will be already handled.

adamaltman: How will they use that info post facto? I don't understand the use case.

slavcodev: I guess this info is used in UI, and perhaps will be used in custom events tracking

adamaltman: Then I suppose the updates suggested are a good idea.

vova07:
retest this please

PR #46, opened 2016-05-26
Branches preview

RomanGotsiy:
Adding possibility to preview spec version from any specific branch as suggested by @slavcodev 

Details:
- to preview specific branch you add the following to the URL: `/preview/<branch-name`, e.g.: [http://rebilly.github.io/RebillyAPI/preview/branches-preview/](http://rebilly.github.io/RebillyAPI/preview/branches-preview/)
- branch gets available for preview after Travis runs on it. Travis won't run deployment on a branch if it doesn't contain appropriate `.travis.yml`. So for old branches, you will not see any previews unless you merge these changes into
- any change made on the branch gets available for preview not immediately but after TravisCI run
- currently, there is no mechanism to clean up old branches. I will either clean up them manually from times to times or set up handler of GitHub hook

Why I didn't use just query param to specify branch (e.g. `?preview=<branch-name>`):
- it requires routing code in the production (master) version
- it is not so flexible: with current approach we are able to have different versions of ReDoc on different previews

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,14 @@ language: node_js
 node_js:
   - "node"
 deploy:
-  skip_cleanup: true
-  provider: script
-  script: npm run deploy
-  on:
-    branch: master
+  - skip_cleanup: true
+    provider: script
+    script: npm run deploy
+    on:
+      branch: master
+  - skip_cleanup: true
+    provider: script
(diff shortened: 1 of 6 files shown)


RomanGotsiy:
I believe it's ready for merge

PR #49, opened 2016-05-30
Add bin and last4 fields to payment card resource response

XaosSintez:


spec/definitions/PaymentCard.yaml
```diff
@@ -12,6 +12,14 @@ properties:
   pan:
     description: The card PAN (Primary Account Number). Required if card is creating not from Token
     type: string
+  bin:
+    description: The card bin (first 6 digits of payment card's number)
```
Zaphell: The Bank Identification Number, usually the PAN's first 6 digits

spec/definitions/PaymentCard.yaml
```diff
@@ -12,6 +12,14 @@ properties:
...
+    type: string
+    readOnly: true
+  last4:
+    description: The payment card's last 4 digits
```
Zaphell: The PAN's last 4 digits

PR #52, opened 2016-06-01
Separate webhook resource

maxyengine:
Added WebsiteWebhook definition and path for website/id/webhook

Resolves: GWY-4076

### DIFF
--- a/spec/definitions/WebsiteWebhook.yaml
+++ b/spec/definitions/WebsiteWebhook.yaml
@@ -1,0 +1,15 @@
+type: object
+required:
+  - webHookUrl
+  - webHookUsername
+  - webHookPassword
+properties:
+  webHookUrl:
+    description: Webhook Url (only for POST/PUT)
+    type: string
+  webHookUsername:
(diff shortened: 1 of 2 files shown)


spec/definitions/WebsiteWebhook.yaml
```diff
@@ -0,0 +1,15 @@
+type: object
+required:
+  - webHookUrl
+  - webHookUsername
+  - webHookPassword
+properties:
+  webHookUrl:
+    description: Webhook Url (only for POST/PUT)
```
slavcodev: Seems the description are deprecated.

spec/definitions/WebsiteWebhook.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - webHookPassword
+properties:
+  webHookUrl:
+    description: Webhook Url
+    type: string
+  webHookUsername:
+    description: Webhook HTTP Authentication Username
+    type: string
+  webHookPassword:
+    description: Webhook HTTP Authentication Password
```
adamaltman: Add to the description:  An empty string will be returned in the response.

PR #53, opened 2016-06-02
Update API docs for Stripe and Paypal gateway accounts adding feature

dkotin:
Resolves: GWY-4145

### DIFF
--- a/spec/definitions/GatewayAccount.yaml
+++ b/spec/definitions/GatewayAccount.yaml
@@ -18,6 +18,17 @@ properties:
   gatewayConfig:
     type: object
     description: Gateway config. Fields depends on gatewayName
+  status:
+    description: The gateway's status
+    readOnly: true
+    type: string
+    enum:
+      - active
+      - inactive
+      - pending
+  redirectUrl:
+    description: This property is write only. Required only for POST and only when adding PayPal or Stripe gateways. Defines the client side location that will receive a respose from API.
+    type: string
   merchantCategoryCode:
     description: The gateway's merchant category code
     type: integer
--- a/spec/definitions/GatewayAccount.yaml
+++ b/spec/definitions/GatewayAccount.yaml
@@ -26,9 +26,6 @@ properties:
       - active
       - inactive
       - pending
-  redirectUrl:
-    description: This property is write only. Required only for POST and only when adding PayPal or Stripe gateways. Defines the client side location that will receive a response from API.
-    type: string
   merchantCategoryCode:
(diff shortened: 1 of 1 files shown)


spec/definitions/GatewayAccount.yaml
```diff
@@ -18,6 +18,17 @@ properties:
...
+  status:
+    description: The gateway's status
+    readOnly: true
+    type: string
+    enum:
+      - active
+      - inactive
+      - pending
+  redirectUrl:
+    description: This property is write only. Required only for POST and only when adding PayPal or Stripe gateways. Defines the client side location that will receive a respose from API.
```
XaosSintez: 1. Typo in word 'respose'
2. Why this write only? I'm sure there's a reason, i just don't know it :-)

adamaltman: I agree with the question about why it's write-only?  

Also, we need to redo this completely to use a discriminator based on the gateway name.  It affects other required fields which are not specified here inside of the `gatewayConfig`.

dkotin: write only because 
```
for POST and only when adding PayPal or Stripe gateways. Defines the client side location that will receive a respose from API.
```
User would never need that. This is the endpoint which API will hit on gateway adding finished.

dkotin: @adamaltman rewrite right in this PR or do that later?

dkotin:
Any other comments? Or lets merge this? @adamaltman @XaosSintez

spec/definitions/GatewayAccount.yaml
```diff
@@ -18,6 +18,17 @@ properties:
...
     description: Gateway config. Fields depends on gatewayName
+  status:
+    description: The gateway's status
+    readOnly: true
+    type: string
+    enum:
+      - active
+      - inactive
+      - pending
+  redirectUrl:
```
slavcodev: wonder, if this is only for concrete gateways, can we write in `gatewayConfig`?

dkotin: Yes, redirect URL can be moved to gatewayConfig.

slavcodev: Really I don't familiar w/ this API and definition. I don't know. I think it make sense.

adamaltman: It might, but we need to implement the use of discriminators for this resource anyway.

dkotin:
@adamaltman lets merge this?

PR #60, opened 2016-06-17
Add ScheduleMethod strategies

slavcodev:


spec/definitions/Scheduling/README.md
```diff
@@ -0,0 +1,19 @@
+# Scheduling objects
+
+## Objects
+
+### Schedule method
+
+Strategies how to calculate relative time. It answer on question "when to schedule"
```
adamaltman: `It answers the question of "when" to schedule.`

vova07:
What about `event-time` strategy ?

spec/definitions/Scheduling/README.md
slavcodev: Thanks.

slavcodev:
> What about event-time strategy ?

That not answer on "when", it answer the question "what is base time"

PR #62, opened 2016-06-27
Support getting GatewayName and Acquirer in Transaction GET API call …

dkotin:
Resolves: GWY-3885


XaosSintez:
I think those fields should be returned always, but with `null` value for other payment methods, for consistency

PR #68, opened 2016-07-01
Create website and add default PaymentGateway on signup (GWY-3690)

dkotin:
Resolves: GWY-3690

### DIFF
--- a/spec/definitions/Signup.yaml
+++ b/spec/definitions/Signup.yaml
@@ -6,7 +6,7 @@ required:
   - lastName
   - businessPhone
   - password
-  - url
+  - website
 properties:
   email:
     description: The user email
--- a/spec/definitions/Signup.yaml
+++ b/spec/definitions/Signup.yaml
@@ -29,12 +29,18 @@ properties:
     description: The user password
     type: string
     format: password
-  url:
-    description: The user's website domain address
+  website:
+    description: The user's website address
     type: string
   currencies:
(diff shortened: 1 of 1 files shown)


spec/definitions/Signup.yaml
```diff
@@ -28,3 +29,12 @@ properties:
     description: The user password
     type: string
     format: password
+  url:
+    description: The user's website domain address
+    type: string
+  currencies:
```
XaosSintez: You must define type of array elements

spec/definitions/Signup.yaml
```diff
@@ -28,3 +29,12 @@ properties:
     description: The user password
     type: string
     format: password
+  url:
```
XaosSintez: May be better not `url`, but something like `website` or `domain` ?

slavcodev: +1 for `website`

spec/definitions/Signup.yaml
```diff
@@ -28,3 +29,12 @@ properties:
...
+    description: The user's website domain address
+    type: string
+  currencies:
+    description: An array of ISO 3-letter currencies codes. Default: ['USD']
+    type: array
+  merchantCategoryCode:
+    description: An array of merchant cateogries ids. Default: 5734 (Computer Software Stores)
+    type: array
```
slavcodev: Also need a items spec

XaosSintez:
:+1:

PR #69, opened 2016-07-04
Review API specs and fix non-consistent APIs

XaosSintez:


spec/definitions/GatewayAccount.yaml
```diff
@@ -87,18 +87,19 @@ properties:
...
-        - Maestro
-        - Solo
-        - Electron
-        - JCB
-        - Voyager
-        - Diners Club
-        - Switch
-        - Laser
+        - visa
+        - master-card
```
adamaltman: I think this is still one word, so without the dash... however, I wonder about this, since they are brand names, should we not use the proper names?

slavcodev: I agreed, We should keep the brand names, original titles.

XaosSintez: So, leave those values as they were before?

adamaltman: Yes

spec/definitions/SubscriptionCancel.yaml
```diff
@@ -6,8 +6,8 @@ properties:
...
     type: string
     enum:
-      - AT_NEXT_RENEWAL
-      - NOW_WITH_PRORATA_REFUND
-      - NOW_WITH_PRORATA_CREDIT
-      - NOW_WITHOUT_REFUND
-      - NOW_WITH_FULL_REFUND
+      - at-next-renewal
+      - now-with-prorata-refund
+      - now-with_prorata-credit
```
Teknologica: Underscore value here, I guess a typo?

spec/definitions/GatewayAccount.yaml
```diff
@@ -99,6 +99,7 @@ properties:
         - Diners Club
         - Switch
         - Laser
+        - China Unionpay
```
2e3s: Don't forget to add Ochapay later

spec/definitions/SubscriptionCancel.yaml
XaosSintez: Yes, thank you

PR #73, opened 2016-07-05
Payment methods missing from SDK for payment api (GWY-4218, GWY-4182)

maxyengine:
Adds missing payment methods

Resolves: GWY-4218, GWY-4182

### DIFF
--- a/spec/definitions/Method.yaml
+++ b/spec/definitions/Method.yaml
@@ -1,8 +1,9 @@
 type: string
 description: The payment method
 enum:
-   - payment_card
-   - paypal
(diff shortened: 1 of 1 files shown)


slavcodev:
This PR affects #69, @XaosSintez please pay attention on this.

XaosSintez:
It's ok i think. I will change payment method values after backend refactor. But i think we should merge this only after backend

slavcodev:
OK, I going to merge this. The backend support these methods at the moment. When backend will be changed to consistent constants, we will have to change spec too

PR #77, opened 2016-07-12
Dunning instruction

slavcodev:


adamaltman:
Github was down.. which is why it didn't pass.

PR #78, opened 2016-07-19
Add HATEOAS links (Customer resource)

XaosSintez:


spec/definitions/Links/DefaultCardLink.yaml
```diff
@@ -0,0 +1,13 @@
+type: object
```
slavcodev: Wouldn't better to use composition for link, to avoid duplication of `href` for example?

spec/definitions/Links/DefaultCardLink.yaml
```diff
@@ -0,0 +1,13 @@
...
+properties:
+  rel:
```
slavcodev: Do you check possibility to use `discriminator` for this attribute?

XaosSintez: Yes, it is still not final version :) Just wanted to know if this solution will work at all and it is the thing we wanted

slavcodev: OK, [there is alt version](https://github.com/Rebilly/RebillyAPI/pull/79)

spec/definitions/Links/Definitions/self.yaml
```diff
@@ -0,0 +1,11 @@
+allOf:
+  - $ref: "#/definitions/CustomerLink"
+  - $ref: "#/definitions/WebsiteLink"
+  - type: object
+    properties:
+      href:
+        description: The resourse permanent link
```
adamaltman: `resource`

spec/definitions/Links/Definitions/self.yaml
```diff
@@ -0,0 +1,11 @@
+allOf:
+  - $ref: "#/definitions/CustomerLink"
+  - $ref: "#/definitions/WebsiteLink"
```
IvanGoncharov: It's not a scalable solution every resource should be mentioned here.
Could you please evaluate https://github.com/IvanGoncharov/RebillyAPI/commit/7f7db26ddb013cd2eee02e1b7113be01fa9cbeef as possible solution?

slavcodev: I'm quite understand the enhancement. Could you please update more links and resources?

slavcodev: > It's not a scalable solution every resource should be mentioned here.

I think it's fine, only `self` link exists in all resource. I do not see how proposed solutions will work.

IvanGoncharov: > Could you please update more links and resources?

I change description for Subscription resource.
https://github.com/IvanGoncharov/RebillyAPI/commit/c4b0270a15128f89930405ac8995b277f3294689

> I think it's fine, only self link exists in all resource.

It makes duplication less of a problem but still you create a lot of objects with `discriminator: rel`.

slavcodev: I still don't understand how it works. Let merge it with this, to be able to see docs.
Could you open PR?

IvanGoncharov:
We discuss this issue together with @RomanGotsiy and have some concerns about how these changes would be represented in documentation. Current approach with discriminator results in the following UI :
image
Dropdown with discriminator values looks confusing. It is even more confusing when you switch discriminator values and only `href` description is changed.

IMHO, the previous version with an array as `items` value and `enum` with one element can be represented significantly better.
In ReDoc we can add special UI element for `enum` with one element and it would be obvious that `rel` is always e.g. `self` in this particular element of the tuple.
Plus if you sort links in a way that optional links go after required ones, you can use `minItems` to mark required links both for ReDoc and tests.

spec/definitions/Links/Definitions/self.yaml
IvanGoncharov: I have some concerns about usage of descriminator: https://github.com/Rebilly/RebillyAPI/pull/78#issuecomment-233894849
But if you decide to stick with descriminator approach we can discuss how to implement it best and I will create this PR.

slavcodev:
The idea with discriminator is to change `href` description, and may be example to show the `URL template`. `minItems` can be with discriminator too to enable constraint, but it confuses me, I think all links a required, so should we also add `maxItems` equals to `minItems`?

Besides some HATEOAS links may include other attributes (type, hreflang, etc)

RomanGotsiy:
The version with tuples allows better visual representation without need for any custom vendor extensions. We have brainstormed with Ivan and for example we can represent tuple like this:
image

@slavcodev what do you think?

slavcodev:
@RomanGotsiy looks pretty nice 👍 

But I wonder how looks definition with tuples. That what I see with `enum` looks strange for me.
If you mean that `CustomerLinks` looks like

~~~yaml
_links:
  type: array
  items:
    - $ref: "#/definitions/SelfLink"
    - $ref: "#/definitions/NotesLink"
    - $ref: "#/definitions/CustomerDefaultCardLink"
  additionalItems: false
  uniqueItems: true
  minItems: 3
  maxItems: 3
~~~

with definition for every link, with concrete "rel", with example of href and possibly additional attributes like `hreflang`, `type`, I'm fine with it.

RomanGotsiy:
Yes. I don't like enum with one value either but  it is used everywhere.
So `items` will look exactly how you specified, but each of definitions will contain `enum` with one value, e.g.:

```yaml
SelfLink:
  type: object
  properties:
    rel:
      type: string
      enum: ["self"]
    href:
      type: string
      description: The resource permanent link
```

btw, links order should be the same as specified in the schema (that's how tuple works)

slavcodev:
> btw, links order should be the same as specified in the schema (that's how tuple works)

yes, and this is great, the serialization on backend always return ordered list.

RomanGotsiy:
Great!
Take into acccount that currently such format is not supported by ReDoc. Most likely I will have it implemented next week.
Reference issues: https://github.com/Rebilly/ReDoc/issues/70, https://github.com/Rebilly/ReDoc/issues/69

RomanGotsiy:
@slavcodev when are you going to merge this? I have to make changes in ReDoc and I don't want to block development.

slavcodev:
@RomanGotsiy it's in `wip`, so I don't review it.
@XaosSintez is this PR ready to review and merge?

XaosSintez:
I will make it mergeable tomorrow, but for `Customer` only. Other resources i will add as separate PRs

PR #83, opened 2016-07-25
Alternative schema of AmountAdjustment

slavcodev:


vova07:

spec/definitions/Dunning/Policies/discount.yaml
```diff
@@ -2,8 +2,12 @@ allOf:
   - $ref: "#/definitions/AmountAdjustment"
   - type: object
     properties:
-      rate:
-        description: The rate of payment amount
-        type: string
+      value:
+        description: The payment amount discount
+        type: integer
```
slavcodev:
~~~
type: number
format: float
~~~

?

vova07:

XaosSintez:
It looks ok for me

vova07:
Hope some of native Enghilsh speakers will can propose a better name for `fixed` attribute.
The idea of this attribute is to tell the merchant that if the value is `true` the system will consider it as `float value` if `false` it will be used as `percent value`. 
Unfortunately my english it's not good enough to propose something better than `fixed`.

PR #96, opened 2016-08-17
Change tags for Histogram endpoint

viceromag:
Current documentation duplicates Histogram endpoint description under Report and Transaction tags / sections. It was decided to use the term Histogram as the most accurate and get rid of duplicates.

### DIFF
--- a/spec/paths/histograms@transactions.yaml
+++ b/spec/paths/histograms@transactions.yaml
@@ -1,8 +1,6 @@
 get:
   tags:
-    - Transaction
     - Histogram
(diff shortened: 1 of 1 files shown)


spec/paths/histograms@transactions.yaml
```diff
@@ -1,8 +1,6 @@
 get:
   tags:
-    - Transaction
     - Histogram
```
adamaltman: I believe this should be changed to Histograms everywhere to be consistent with the plural naming of tags.

PR #97, opened 2016-08-17
Attachments and Files specs (GWY-4148)

dkotin:
Resolves: GWY-4148


spec/definitions/Attachment.yaml
```diff
@@ -0,0 +1,56 @@
...
+  - relatedId
+properties:
+  id:
+    description: Unique Attachment ID
+    type: string
+    readOnly: true
+  fileId:
+    description: Linked File object id
+    type: string
+    readOnly: true
```
slavcodev: Typo?

dkotin: no. Also your input regarding File POST  greatly appreciated

PR #102, opened 2016-08-30
Reset Sandbox

XaosSintez:


spec/definitions/ResetSandbox.yaml
```diff
@@ -0,0 +1,12 @@
+type: object
+properties:
+  createdTime:
```
vova07: what about `startTime` or something like that instead ?

spec/paths/reset-sandbox.yaml
```diff
@@ -0,0 +1,23 @@
+post:
+  tags:
+    - ResetSandbox
+  summary: Reset sandbox
+  description: |
+    Reset sandbox
+  responses:
+    201:
+      description: Sandbox was reseted
```
slavcodev: I believe this operation should be async, It can last too long.

XaosSintez: It is already async as Adam wrote in backend's PR

slavcodev: It maybe async in backend, but my question relates to API, for async operation server should responded with `202 ACCEPTED`, because server do not know when operation will finish. Also server should redirect the client to queued item, the client can track when the operation ends.

2e3s: "Sandbox was reset" I guess

slavcodev: @2e3s what do you mean?

adamaltman: He means "reseted" isn't a real word, but "reset" is.  `Sandbox was reset` :+1:  vs. `Sandbox was reseted` :-1:

spec/paths/reset-sandbox.yaml
```diff
@@ -0,0 +1,23 @@
+post:
+  tags:
+    - ResetSandbox
```
adamaltman: If this does belong in our REST API, it seems strange to give it its own tag, which has equal weight as something like Payments.  Maybe this should go into a `Utilities` tag.  If we keep Reset Sandbox, then the tag should be `Reset Sandbox` (with a space).  However, `Reset Sandbox` and `Migrate Payment Cards` (and `Status`) all seem like `Utilities` to me, and not so much like RESTful resources.

Teknologica:
A month later, has there been any update to this?

slavcodev:
This seems is not going to be merged.

PR #103, opened 2016-08-30
2FA functionality in user profile module (GWY-4409)

dkotin:
Implements renewing for totpSecret

Resolves: GWY-4409


XaosSintez:
I wonder if we use `case-like-this` in urls? `/renewTotpSecret` looks ugly

XaosSintez:
:+1:

PR #104, opened 2016-08-30
Update API specs related with dunning renaming (GWY-4415)

vova07:


spec/definitions/Payment.yaml
```diff
@@ -46,8 +46,10 @@ properties:
     type: object
     description: Dunning instructions object for payment declines
     properties:
+      # TODO: Rename to paymentRetry when Payment will be refactored
       dunning:
-        $ref: "#/definitions/Dunning"
+        $ref: "#/definitions/PaymentRetry"
+      # TODO: Rename to paymentRetryIndex when Payment will be refactored
```
slavcodev: Could we use `retryCount` instead of `paymentRetryIndex`?

slavcodev:
👍

PR #108, opened 2016-08-31
Add Rules Tracking API spec (GWY-4137)

vova07:


spec/definitions/Tracking/RulesTracking.yaml
```diff
@@ -0,0 +1,36 @@
...
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  eventTime:
+    description: The time when event has happened
+    allOf:
+      - $ref: "#/definitions/ServerTimestamp"
+  rulesVersion:
+    type: integer
+    description: The rules version
+  criteriaExcerpt:
```
Teknologica: What's the need behind this? As opposed to looking at the content of `rules` directly?

spec/definitions/Tracking/RulesTracking.yaml
```diff
@@ -0,0 +1,36 @@
+type: object
+description: Tracking Rules.
+readOnly: true
+properties:
+  eventType:
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  eventTime:
```
slavcodev: I suggest to use `occurredTime` or `createdTime` instead, I don't know, but `eventTime` confusing me, also to be consistent, we use in our API `verb in past form` + `Time` in other places.

spec/paths/tracking@rules@{id}@{version}.yaml
```diff
@@ -0,0 +1,23 @@
+parameters:
```
slavcodev: I don't think we need to implement this endpoint. Seems make no sense for that. We can filter `tracking/rules/{id}?filter=rulesVersion:1`

spec/definitions/Tracking/RulesTracking.yaml
slavcodev: @Teknologica, `rules` here a list of executed rules, and result if it matched or not, w/o it would be painful for merchant to check their rules and manually check criteria to check if rules matched or not.

spec/definitions/Tracking/RulesTracking.yaml
```diff
@@ -0,0 +1,36 @@
...
+    description: The time when event has happened
+    allOf:
+      - $ref: "#/definitions/ServerTimestamp"
+  rulesVersion:
+    type: integer
+    description: The rules version
+  criteriaExcerpt:
+    type: array
```
slavcodev: It seems to be `object`, not array.

spec/definitions/Tracking/RulesTracking.yaml
adamaltman: I think `createdTime` just for consistency with other resources would be good.

PR #113, opened 2016-09-02
Coupons

XaosSintez:


spec/definitions/Coupon.yaml
```diff
@@ -0,0 +1,68 @@
+type: object
+description: Coupons and Discounts
+properties:
+  id:
+    readOnly: true
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  code:
```
slavcodev: Is not the code unique? Could we use `id` instead of another unique attribute?

spec/definitions/Coupon.yaml
```diff
@@ -0,0 +1,68 @@
...
+        default: 0
+  applicationRestrictions:
+    $ref: "#/definitions/applicationRestrictions"
+  redemptionCount:
+    type: integer
+    description: Coupon's redemption count
+    minimum: 0
+    readOnly: true
+  status:
+    type: boolean
```
slavcodev: `status` better to do string, with enum `active`, `inactive`. We have in rules branch the definition `OnOff`, may be would be useful.

spec/definitions/Coupon.yaml
```diff
@@ -0,0 +1,68 @@
...
+        type: integer
+        minimum: 0
+        default: 0
+      limitPerCustomer:
+        description: Coupon's limit per one Customer. 0 for unlimited
+        type: integer
+        minimum: 0
+        default: 0
+  applicationRestrictions:
+    $ref: "#/definitions/applicationRestrictions"
```
slavcodev: wrong case for `a`

slavcodev:
I suggest to put the definitions into `Coupons` folder.

spec/definitions/Coupon.yaml
2e3s: We should be able to change the code.

slavcodev: Hmm, I do not think so. Since coupon is created and shared, do we can change this code?

XaosSintez: Yes, from requirements:

> As a merchant, I may want to change the redemption code for a coupon, to prevent people sharing something secret that I don’t want shared.

slavcodev: Hmm, what is the difference then to deactivate current coupon and release new one?

XaosSintez: It was new client's 'request' as i understand. I can't say much about `why?`, i just did as i read in requirements. It will not add us many problems, so why not?

slavcodev: N/p, I  just clarifying the requirements. It will not add many problems, but I think make no sense to add meaningless fields in API and support it.

spec/definitions/Coupon/CouponTracking.yaml
```diff
@@ -0,0 +1,37 @@
...
+  id:
+    readOnly: true
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  couponId:
+    description: Coupon's ID
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  resourceType:
+    description: Resoource type
```
colinbird: typo

spec/definitions/Coupon/CouponTracking.yaml
```diff
@@ -0,0 +1,37 @@
...
+    type: string
+    readOnly: true
+    enum:
+      - invoice
+      - subscription
+      #- product
```
colinbird: customer?

spec/definitions/Coupon.yaml
slavcodev: @XaosSintez please check the doc comments near that requirement, Adam confirmed that code is immutable after coupon is issued.

spec/definitions/InvoiceItem.yaml
```diff
@@ -32,6 +32,11 @@ properties:
     description: End time
     type: string
     format: date-time
+  coupons:
```
2e3s: Perhaps it would be another type of invoice item "discount" (as debit or credit)

spec/definitions/Coupon/Coupon.yaml
```diff
@@ -52,16 +40,20 @@ properties:
...
     description: Coupon created time
     allOf:
       - $ref: "#/definitions/ServerTimestamp"
   updatedTime:
-    description: Coupon updated time. Coupon can not be updated after first redeem
-    allOf:
-      - $ref: "#/definitions/ServerTimestamp"
-  archivedTime:
-    description: Coupon archived time
+    description: Coupon updated time. Coupon can not be updated after first redeem. Also ths timestamp is changed after every redemption.
```
slavcodev: It's read-only attribute, it can not be updated ever. I think the description of this attribute should be changed, this is system attribute and it's the same for all resources.

spec/definitions/Coupon/CouponTracking.yaml
```diff
@@ -0,0 +1,37 @@
...
+properties:
+  id:
+    readOnly: true
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  couponId:
+    description: Coupon's ID
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  resourceType:
```
adamaltman: I'm not sure about this... isn't a coupon just related to customers (upon redemptions) and invoices (upon applying discounts)?

spec/definitions/Coupon/apply-until.yaml
```diff
@@ -0,0 +1,17 @@
...
+  -
+    type: "object"
+    required:
+     - "value"
+    properties:
+      value:
+        type: string
+        description: Restriction value #todo
+        enum:
+          - single-use
```
adamaltman: I think `single-use` and `invoices-applied-count` should be combined into `fixed-use` (or `fixed-use-count`)... if someone has a value of `1`, that would be a (single-use).  

Also, I think `until-canceled` could be removed, as all coupons can be canceled.  

In fact, I think the only two options we should have are `fixed-uses` and `fixed-period` (and even then, I'm not really sure why `fixed-period` and why not `fixed-uses`).  However, there are some valid edge cases in my head that may make sense for `fixed-period`, it may make more sense to leave `fixed-period` until someone requests that implementation.  So that leaves us with only `fixed-uses`, which means that we could probably change `apply-until` to `apply-until-fixed-uses-count` (or something like that).

spec/definitions/Coupon/CouponRestriction.yaml
```diff
@@ -0,0 +1,17 @@
...
+type: object
+discriminator: "type"
+properties:
+  type:
+    description: Coupon's restriction type
+    type: array
+    items:
+      type: string
+      enum:
+        - apply-until
```
adamaltman: I have a comment below about `apply-until`.  Perhaps just renaming it to `apply-limit-per-redemption`.  (And the value would be a numeric int.)

adamaltman:
How is a coupon redeemed?  I only see `GET` but not `POST`...

spec/definitions/Invoice.yaml
```diff
@@ -39,6 +39,11 @@ properties:
     readOnly: true
     items:
       type: string
+  coupons:
```
2e3s: Here and then: I think it should be `couponIds`

spec/definitions/Coupon/Coupon.yaml
```diff
@@ -0,0 +1,65 @@
...
+required:
+  - type
+  - value
+properties:
+  id:
+    description: Coupon's ID, also it is redemption code
+    readOnly: true
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  type:
```
slavcodev: There would be better, IMO, to group this and next 2 attributes in object `discount` with discriminator, because `currency` are meaningless for `percent`

spec/definitions/Coupon/Coupon.yaml
```diff
@@ -0,0 +1,65 @@
...
+      - fixed
+  value:
+    description: Discount value
+    type: number
+    format: double
+  currency:
+    description: Discount currency. Required for fixed type
+    type: string
+  restrictions:
+    description: Coupon redemption restrictions
```
slavcodev: These restrictions for discounting too, so `redemption restrictions` is confusing.

spec/definitions/Coupon/Coupon.yaml
```diff
@@ -0,0 +1,65 @@
+type: object
+description: Coupons and Discounts
+required:
+  - type
+  - value
+properties:
+  id:
+    description: Coupon's ID, also it is redemption code
```
slavcodev: May be call it `code` or `redeemCode`? Seems is more suitable in context of coupons.

spec/definitions/Coupon/Coupon.yaml
```diff
@@ -0,0 +1,65 @@
...
+    type: integer
+    description: Coupon's redemption count
+    minimum: 0
+    readOnly: true
+  status:
+    type: string
+    description: If coupon enabled
+    readOnly: true
+    enum:
+      - active
```
slavcodev: wouldn't e better to use statuses from `UL`? `issued` and `expired` instead of `active` and `inactive`?

spec/definitions/Coupon/CouponRedemption.yaml
```diff
@@ -0,0 +1,37 @@
...
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  couponId:
+    description: Coupon's ID
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  resourceType:
+    description: Resource type
+    type: string
+    readOnly: true
```
slavcodev: Hmm, is it read-only? How to redeem to concrete resource?

spec/definitions/Coupon/CouponRedemption.yaml
```diff
@@ -0,0 +1,37 @@
+type: object
+description: Coupons redeem log
+properties:
```
slavcodev: Seems missing the `customerId` for redemption.

spec/definitions/Coupon/Coupon.yaml
```diff
@@ -3,9 +3,10 @@ description: Coupons and Discounts
 required:
   - type
   - value
+discriminator: "type"
```
slavcodev: Well, I meant to remove `type` from coupon, coupon does not have the type, coupon have different kinds of discount, IMO. I.e. I've suggested

~~~json
{
  "redeemCode": "foo",
  "discount": {
    "type": "fixed",
    "value": 10000,
    "currency": "USD"
  }
}
~~~

spec/definitions/Coupon/Coupon.yaml
```diff
@@ -37,8 +31,8 @@ properties:
     description: If coupon enabled
     readOnly: true
     enum:
-      - active
-      - inactive
+      - issued
+      - exipred
```
slavcodev: type `expired`

spec/definitions/Coupon/Coupon.yaml
2e3s: not some `value`, particular `amount` then maybe

slavcodev: yes, for `fixed`, the `value` is more suitable, I just copy-pasted from existing specs

spec/definitions/Coupon/fixed.yaml
```diff
@@ -1,19 +1,17 @@
 description: Coupon fixed discount
 allOf:
   -
-    $ref: "#/definitions/Coupon"
+    $ref: "#/definitions/Discount"
   -
     type: "object"
     required:
-     - "discount"
+     - "value"
```
slavcodev: We're usually using `amount` there.

spec/definitions/Coupon/Coupon.yaml
```diff
@@ -0,0 +1,55 @@
...
+  discount:
+    description: Coupon's discount
+    allOf:
+      - $ref: "#/definitions/Discount"
+  restrictions:
+    description: Coupon restrictions
+    type: array
+    items:
+     - $ref: "#/definitions/CouponRestriction"
+  redemptionCount:
```
slavcodev: Wonder, how is correctly `redemptionCount` or `redemptionsCount`? Or may be just `redemptions`?

spec/definitions/Coupon/CouponRedemption.yaml
```diff
@@ -0,0 +1,39 @@
...
+    type: string
+    enum:
+      - invoice
+      - subscription
+      #- product
+  resourceId:
+    description: Resource ID
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  redeemTime:
```
slavcodev: `redeemedTime` ?

slavcodev: Also is missing `canceledTime`

spec/definitions/Coupon/Discount.yaml
```diff
@@ -0,0 +1,9 @@
+type: object
+discriminator: "type"
+properties:
+  type:
+    description: Discount type
+    type: string
+    enum:
+      - percent
```
slavcodev: I'm wonder, should it be `percent` or `percentage`?

spec/definitions/Coupon/Discount.yaml
```diff
@@ -0,0 +1,9 @@
...
+      - fixed
```
slavcodev: We should consider to rename this to `money`

spec/definitions/Coupon/apply-to-products.yaml
```diff
@@ -0,0 +1,14 @@
+description: apply-to-products restrictions
+allOf:
+  -
+    $ref: "#/definitions/CouponRestriction"
+  -
+    type: "object"
+    required:
+     - "value"
+    properties:
+      value:
```
slavcodev: Wouldn't for consistency it should be `productIds` ?

spec/definitions/Coupon/fixed-use-count.yaml
```diff
@@ -0,0 +1,12 @@
+description: fixed-use-count restrictions
+allOf:
+  -
+    $ref: "#/definitions/CouponRestriction"
+  -
+    type: "object"
+    required:
+     - "value"
+    properties:
+      value:
```
slavcodev: How about `quantity` instead of `value`?

spec/definitions/Coupon/restrict-to-plans.yaml
```diff
@@ -0,0 +1,14 @@
+description: restrict-to-plans restrictions
+allOf:
+  -
+    $ref: "#/definitions/CouponRestriction"
+  -
+    type: "object"
+    required:
+     - "value"
+    properties:
+      value:
```
slavcodev: Same here, should it be `planIds` ?

spec/definitions/Coupon/restrict-to-products.yaml
```diff
@@ -0,0 +1,14 @@
+description: restrict-to-products restrictions
+allOf:
+  -
+    $ref: "#/definitions/CouponRestriction"
+  -
+    type: "object"
+    required:
+     - "value"
+    properties:
+      value:
```
slavcodev: `productIds` ?

spec/definitions/Coupon/total-quantity.yaml
```diff
@@ -0,0 +1,12 @@
+description: Total quantity restrictions
+allOf:
+  -
+    $ref: "#/definitions/CouponRestriction"
+  -
+    type: "object"
+    required:
+     - "value"
+    properties:
+      value:
```
slavcodev: Wouldn't `quantity` better ?

spec/definitions/Coupon/CouponRestriction.yaml
```diff
@@ -0,0 +1,17 @@
...
+properties:
+  type:
+    description: Coupon's restriction type
+    type: array
+    items:
+      type: string
+      enum:
+        - fixed-use-count
+        - minimum-order-value
+        - apply-to-products
```
slavcodev: Having problems to understand difference between `apply-to-products` and `restrict-to-products`. Can anybody help to understand?

spec/definitions/Coupon/CouponRestriction.yaml
```diff
@@ -0,0 +1,17 @@
...
+type: object
+discriminator: "type"
+properties:
+  type:
+    description: Coupon's restriction type
+    type: array
+    items:
+      type: string
+      enum:
+        - fixed-use-count
```
slavcodev: I believe this value is confusing, there are some alternative values `discounts-per-customer`, `discount-quantity`, `discounts-per-redemptions`, etc.

spec/definitions/Coupon/Coupon.yaml
2e3s: It should have `count` for clarity. Not sure about the first thing though. `redemptionCount` feels artificial.

spec/definitions/Coupon/CouponRestriction.yaml
XaosSintez: In Coupons document, at the end of page 15:

> restrict-to-plans, restrict-to-products (different from Apply in that it is a Redemption restriction, not an application restriction).

But also there's

> Type: apply-to-products  DOES NOT EXIST IN REBILLY YET

So may be we should remove `apply-to-products` for now?

spec/definitions/Coupon/Discount.yaml
XaosSintez: I think `percent` is better here

spec/definitions/Coupon/CouponRedemption.yaml
```diff
@@ -24,11 +24,16 @@ properties:
     description: Resource ID
     allOf:
       - $ref: "#/definitions/ResourceId"
-  redeemTime:
+  redeemedTime:
     description: Coupon redeem time
     readOnly: true
     allOf:
       - $ref: "#/definitions/ServerTimestamp"
+  redeemedTime:
```
slavcodev: I guess you meant `canceledTime`

spec/definitions/Coupon/Discount.yaml
slavcodev: I don't think so. `percent` is noun, `percentage` and `fixed` are adjectives.
so 
- or coupon provides discount: `percent` or `money`
- or coupon provides `percentage` or `fixed` discount

does that make sense?

XaosSintez: I don't mind to rename `fixed`, but i don't like `money`. May be `amount`?

spec/definitions/Coupon/CouponRestriction.yaml
slavcodev: I guess this is duplicate and appear because of 2 separate groups of restrictions in the past (redeemedRestrictions, applicationRestrictions), but now when we combine it, I'm pretty sure it is the same.

spec/definitions/Coupon/discounts-per-redemption.yaml
```diff
@@ -0,0 +1,12 @@
+description: fixed-use-count restrictions
```
adamaltman: Change this description to be Discounts per redemption restriction.

spec/paths/coupons@{id}.yaml
```diff
@@ -0,0 +1,68 @@
...
+  description: |
+    Create or update a coupon with predefined identifier string
+  parameters:
+    - name: body
+      in: body
+      description: Coupon resource
+      required: true
+      schema:
+        $ref: "#/definitions/Coupon"
+  responses:
```
adamaltman: What if the coupon can't be updated, due to a conflict (like if it has already been used, so you can't modify the percentage discount after that)?

spec/definitions/Coupon/CouponRedemption.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - $ref: "#/definitions/ResourceId"
+  couponId:
+    description: Coupon's ID
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  customerId:
+    description: Customer's ID
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  resourceType:
```
adamaltman: I understood this differently.  For example, there would be a member:  `additionalRestrictions` which would have an array of addition `CouponRestriction` data.  Also, I think the `CouponRestriction` data is missing two simple restrictions:  `restrict-to-invoiceIds`, `restrict-to-subscriptionIds`...

spec/definitions/Coupon/Discount.yaml
adamaltman: I think it should be `percent`.  For the second one, it takes a type `Money` as the input (currency and amount).  But I don't think that is the discount type.  The discount type is `fixed amount`, or `fixed`.

Also, `percentage` is a noun: http://www.dictionary.com/browse/percentage (and percent is a noun or adjective: http://www.dictionary.com/browse/percent).  For example, you would say:  I got a 10 percent discount (not I got a 10 percentage discount).

spec/definitions/Coupon/CouponRedemption.yaml
adamaltman: That has been addressed.  It seems to be missing the `additionalRestrictions` array.

spec/paths/coupons-redemptions@{id}.yaml
```diff
@@ -0,0 +1,15 @@
+parameters:
+  - $ref: "#/parameters/resourceId"
+get:
```
adamaltman: Can you filter by a `couponId`?  also, by any other parameters, like `customerid`?

spec/definitions/Coupon/CouponRedemption.yaml
```diff
@@ -0,0 +1,38 @@
+type: object
+description: Coupons redeem log
+properties:
+  id:
+    readOnly: true
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  couponId:
```
slavcodev: `couponRedeemCode` ?

spec/definitions/Coupon/CouponRedemption.yaml
```diff
@@ -0,0 +1,38 @@
+type: object
+description: Coupons redeem log
+properties:
+  id:
+    readOnly: true
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  couponRedeemCode:
+    description: Coupon's ID
```
2e3s: If this is a coupon identifier, then how can a merchant use a code again?

adamaltman:
We need to declare the `Coupons` tag in the `swagger.yaml` preferably in the appropriate alphabetical order.

spec/definitions/Coupon/Coupon.yaml
```diff
@@ -0,0 +1,54 @@
...
+required:
+  - discount
+properties:
+  redeemCode:
+    description: Coupon's redemption code
+    readOnly: true
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  discount:
+    description: Coupon's discount
```
slavcodev: This need to remove, by JSON-Schema spec we cannot use another properties together with `$ref`.

PR #114, opened 2016-09-05
Transactions: add response code and response message (GWY-4425)

dkotin:
Resolves: GWY-4425

### DIFF
--- a/spec/definitions/Transaction.yaml
+++ b/spec/definitions/Transaction.yaml
@@ -61,6 +61,14 @@ properties:
     description: The transactions's processor response
     type: string
     readOnly: true
+  gatewayResponseCode:
+    description: The gateway's response code
+    type: string
+    readOnly: true
+  gatewayResponseMessage:
+    description: The gateway's response message
+    type: string
+    readOnly: true
   websiteId:
     description: Website's ID
     allOf:
--- a/spec/definitions/Transaction.yaml
+++ b/spec/definitions/Transaction.yaml
@@ -52,27 +52,60 @@ properties:
     allOf:
       - $ref: "#/definitions/ResourceId"
     readOnly: true
-  gatewayTransactionId:
-    description: The gateway's transaction ID
-    allOf:
-      - $ref: "#/definitions/ResourceId"
+  gateway:
+    type: object
+    description: The related gateway information
     readOnly: true
-  gatewayCode:
-    description: The gateway's response code
-    type: string
-    readOnly: true
-  gatewayMessage:
-    description: The gateway's response message
-    type: string
-    readOnly: true
-  originalGatewayCode:
-    description: The raw, unmapped gateway's response code
-    type: string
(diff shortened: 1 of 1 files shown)


slavcodev:
Sorry, on Friday the names seems ok, but today I think `gatewayCode` a bit confusing, not clear what means "code of gateway".

So to decide there, we need to consider some things:

- We're having transaction processed through gateway, which have response and other which don't, like `cash transactions`
- We're missing a lot of gateway response info: code, message, AVS-response, CVV-response

How about optional object?

~~~json
{
 "gateway": {
  "transactionId": "...",
  "responseCode": "...",
  "responseMessage": "...",
  "responseOriginalCode": "...",
  "responseOriginalMessage": "...",
  "avsResponseCode": "...",
  "avsResponseMessage": "...",
  "cvvResponseCode": "...",
  "cvvResponseMessage": "..."
 }
}
~~~

or more structural

~~~json
{
 "gateway": {
  "transactionId": "...",
  "response": {
   "code": "...",
   "message": "...",
  "originalCode": "...",
  "originalMessage": "..."
  },
  "avsResponse": {
   "code": "...",
   "message": "..."
  },
  "cvvResponse": {
   "code": "...",
   "message": "..."
  }
 }
}
~~~

spec/definitions/Transaction.yaml
```diff
@@ -52,23 +52,60 @@ properties:
...
+  gateway:
+    type: object
+    description: The related gateway information
+    readOnly: true
+    properties:
+      transactionId:
+        description: The gateway's transaction ID
+        allOf:
+          - $ref: "#/definitions/ResourceId"
+        readOnly: true
```
slavcodev: `gateway` is fully read-only, make no sense to add this for internal properties.

spec/definitions/Transaction.yaml
```diff
@@ -52,23 +52,60 @@ properties:
...
-  gatewayResponseMessage:
-    description: The gateway's response message
-    type: string
-    readOnly: true
+  gateway:
+    type: object
+    description: The related gateway information
+    readOnly: true
+    properties:
+      transactionId:
```
slavcodev: To do not break compatibility, I suggest let's leave it as is `gatewayTransactionId`, we can move it in `gateway` later when we will be fully agree with this solution.

dkotin: I suggest to leave that in the code while remove in the doc. so it would be still compatible but new users will not use that.

XaosSintez: Yes, i think that's the way we usually do

slavcodev: Yes, we do in that way, in case we decided to change. In case  with `gatewayTransactionId`, was just proposal. It's definitely not a part of this ticket, which we need to finish asap.

adamaltman:
:+1:

PR #115, opened 2016-09-06
Add a safe hash to PaymentCard model

viceromag:
Related to GWY-4437


spec/definitions/PaymentCard.yaml
```diff
@@ -36,6 +36,10 @@ properties:
   token:
     description: PaymentCardToken. Use without any other fields
     type: string
+  safeHash:
```
adamaltman: We shall need discuss this offline...

PR #118, opened 2016-09-07
Update credential spec

vova07:


spec/definitions/Rules/WebhookAuthorization/EncryptedAuthorization.yaml
```diff
@@ -0,0 +1,7 @@
+type: object
+properties:
+  token:
+    type: string
+    description: The credential identification token
+required:
+  - token
```
slavcodev: New line needed here

PR #127, opened 2016-09-15
Add Webhook Credential Hashes spec

vova07:


slavcodev:
wonder what is better endpoint `/credential-hashes/webhook/` or `/credential-hashes/webhooks/` , thoughts?

adamaltman:
I think with `/webhooks/`.

spec/definitions/WebhookCredentials/WebhookAuthorization.yaml
```diff
@@ -0,0 +1,23 @@
+type: object
+discriminator: type
+properties:
+  token:
+    readOnly: true
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  host:
```
adamaltman: is there no port too?  Is host just the base domain, or the full url?  Example:

1. https://example.com
2. example.com
3. https://example.com/receive/webhook/here

slavcodev: It's host (no scheme, no path), just to link credential by host, the port also no need here. 
To be clear, it's not webhook URL. The domain from webhook URL will be compare with this host before allow to use credential.

adamaltman: Okay, so it's like 2. `example.com`.  The logic makes sense.

slavcodev: Yes. Only domain/host `example.com`. May be some description and example would be helpful.

spec/paths/credential-hashes@webhooks@{token}.yaml
```diff
@@ -0,0 +1,29 @@
+parameters:
+  -
+    name: token
```
slavcodev: `hash` ?

vova07: fixed!

PR #135, opened 2016-09-26
Change spec and sdk for api log user loggined (GWY-4553)

maxyengine:
Add TrackingUser

Resolves: GWY-4553

### DIFF
--- a/spec/definitions/Tracking/ApiTracking.yaml
+++ b/spec/definitions/Tracking/ApiTracking.yaml
@@ -30,6 +30,9 @@ properties:
   response:
     type: string
     description: Response JSON-string
+  user:
+    allOf:
+      - $ref: "#/definitions/TrackingUser"
   duration:
     type: integer
     description: Request duration in milliseconds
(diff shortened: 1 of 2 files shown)


spec/definitions/Tracking/ApiTracking.yaml
```diff
@@ -30,6 +30,24 @@ properties:
   response:
     type: string
     description: Response JSON-string
+  user:
```
adamaltman: why not the user id?

maxyengine: I think this is unnecessary information for the merchant, in case when user is admin

adamaltman: What do you mean by `admin`?  Rebilly support user? Or something else?

maxyengine: support user

maxyengine: added more fields

adamaltman:
:+1:

PR #148, opened 2016-10-18
Add SendTestEmail spec

vova07:


vova07:

spec/swagger.yaml
```diff
@@ -246,6 +246,9 @@ tags:
       website is related to each invoice and each payment gateway account.  This feature
       would allow you to have gateway accounts that are related to multiple websites, or
       exclusive to particular websites.  And gives you more control over your business.
+  - name: Send Test Email
```
adamaltman: It seems strange for this to occupy it's own first level navigation in our API docs

vova07: Rename to `Test Rule Actions`

adamaltman:
What do you think about a path `previews` instead of `tests`?  To me, tests just seems more like unit tests, etc...  Whereas, this functionality is more like "try it out".

vova07:
Yes, it seems better name.

vova07:

spec/paths/previews@rule-actions@send-email.yaml
```diff
@@ -0,0 +1,33 @@
+post:
+  tags:
+    - Test Rule Actions
```
adamaltman: The tag should be adjusted too.  I think it may be best for it to be `Rules` as this may be better to fall within that in the left-nav.  Already the left-nav is overwhelming

vova07: Done!

PR #149, opened 2016-10-21
Fix gateway account names

XaosSintez:


slavcodev:
@XaosSintez Could you please elaborate reverting names?
AFAIK, we've decided to use brand-names,
and for example it is `OchaPay`, is not `Ochapay`.

oliviertr:
Related to: https://jira.ilabsinc.com/browse/GWY-4768

Could just as well change the names of the gateway definition objects instead though.

XaosSintez:
@slavcodev @oliviertr 
Yes, this was fixed in another PR, thank you all

PR #151, opened 2016-10-22
Transaction DCC

XaosSintez:


spec/definitions/Transaction.yaml
```diff
@@ -134,6 +134,43 @@ properties:
...
+        type: object
+        description: Quote amount and currency
+        properties:
+          amount:
+            type: number
+            format: double
+          currency:
+            type: string
+      usdMarkup:
+        description: TODO
```
adamaltman: description: `The amount of markup translated to USD.`

spec/definitions/Transaction.yaml
```diff
@@ -134,6 +134,43 @@ properties:
...
+    type: boolean
+    readOnly: true
+  dcc:
+    description: Dcc detailed information. Can be null if hasDcc is false
+    type: object
+    readOnly: true
+    properties:
+      base:
+        type: object
+        description: Base amount and currency
```
2e3s: Although this is a normal domain language, I suggest to make the description clearer for anyone:
base - initial amount and currency to convert from
quote - suggested amount and currency to convert to

spec/definitions/Transaction.yaml
```diff
@@ -134,6 +134,43 @@ properties:
     description: Acquirer name, available only when method is 'payment_card', else null
     type: string
     readOnly: true
+  hasDcc:
+    description: True if transaction has DCC applied
+    type: boolean
+    readOnly: true
+  dcc:
+    description: Dcc detailed information. Can be null if hasDcc is false
```
2e3s: I believe `Dynamic Currency Conversion` is better to use in the descriptions than `dcc` which is an abbreviation so should be in capital: `DCC`.
Also the next sentence would be `null if hasDcc is false` (more correct language  I think)

spec/definitions/Transaction.yaml
XaosSintez: @adamaltman thanks, fixed

spec/definitions/Transaction.yaml
```diff
@@ -134,6 +134,43 @@ properties:
...
+        description: Quote amount and currency
+        properties:
+          amount:
+            type: number
+            format: double
+          currency:
+            type: string
+      usdMarkup:
+        description: The amount of markup translated to USD
+        type: integer
```
2e3s: float

XaosSintez: @2e3s okay, thank you for review!

PR #154, opened 2016-10-27
Add cancelSource

colinbird:


spec/definitions/Subscription.yaml
```diff
@@ -44,6 +44,9 @@ properties:
     description: Autopay determines if a payment attempt will be automatic
     type: boolean
     default: true
+  cancelSource:
+    description: Source of cancellation
+    type: string
```
XaosSintez: it should be
`readOnly:  true`

PR #158, opened 2016-11-03
Transactions API missing data found in legacy (GWY-4835)

maxyengine:
Add child transactions to response

Resolves: GWY-4835

### DIFF
--- a/spec/definitions/Transaction.yaml
+++ b/spec/definitions/Transaction.yaml
@@ -53,6 +53,12 @@ properties:
     allOf:
       - $ref: "#/definitions/ResourceId"
     readOnly: true
+  childTransactions:
+    description: The children transactions IDs
+    readOnly: true
+    type: array
+    items:
+      - $ref: "#/definitions/ResourceId"
   rebillNumber:
     description: The transactions's rebill number
     type: integer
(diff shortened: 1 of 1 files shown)


XaosSintez:
Is that all missing data? Or is there something else?

maxyengine:
The rest are missing in spec

spec/definitions/Transaction.yaml
```diff
@@ -134,6 +140,12 @@ properties:
     description: Acquirer name, available only when method is 'payment_card', else null
     type: string
     readOnly: true
+  velocity:
+    description: Transaction's velocity
```
adamaltman: `The number of transactions by the same customer in the past 24 hours`

spec/definitions/Transaction.yaml
```diff
@@ -134,6 +140,12 @@ properties:
...
+    type: number
+    format: float
```
adamaltman: is `integer` not an option?

spec/definitions/Transaction.yaml
maxyengine: thanks

spec/definitions/Transaction.yaml
maxyengine: right, i'll fix it

PR #163, opened 2016-11-10
Make consistent expiredTime

XaosSintez:


XaosSintez:
@slavcodev yes, we decided. I will do it as separate PR

PR #164, opened 2016-11-10
Add webhook preview spec

vova07:


vova07:

spec/paths/previews@rule-actions@trigger-webhook.yaml
```diff
@@ -0,0 +1,33 @@
...
+  summary: Trigger a test webhook
+  description: |
+    Trigger a test webhook
+  parameters:
+    - name: body
+      in: body
+      description: Test webhook resource
+      required: true
+      schema:
+        $ref: "#/definitions/SendPreviewWebhook"
```
Teknologica: So since `SendPreviewWebhook` is an alias for `Webhook`, we can specify a new URI and authorization. Likely the user would keep the default values (from the action) and be given the option to modify it for testing only. That's the correct assumption right?

Teknologica:
```js
{
    body: { 
        method,
        query,
        uri,
        body,
        auth,
        headers
    }
}
```

vova07: Yes, it's depends on how it will be implemented in `frontend`. You can use the default values, or you can modify that values depending on what is more logical for `UI` and the end users.

PR #169, opened 2016-11-16
Add spec for invoiseIds and subscriptionIds (GWY-4856)

maxyengine:
Resolves: GWY-4856

### DIFF
--- a/spec/definitions/Transaction.yaml
+++ b/spec/definitions/Transaction.yaml
@@ -59,6 +59,18 @@ properties:
     type: array
     items:
       - $ref: "#/definitions/ResourceId"
+  invoiceIds:
+    description: The invoices related to transaction
+    readOnly: true
+    type: array
+    items:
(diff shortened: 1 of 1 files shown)


spec/definitions/Transaction.yaml
```diff
@@ -59,6 +59,18 @@ properties:
     type: array
     items:
       - $ref: "#/definitions/ResourceId"
+  invoiceIds:
+    description: The invoices related to transaction
+    readOnly: true
+    type: array
+    items:
+      - $ref: "#/definitions/ResourceId"
+  subscribeIds:
```
2e3s: subscriptionIds

spec/definitions/Transaction.yaml
```diff
@@ -59,6 +59,18 @@ properties:
...
+    description: The subscribes related to transaction
```
2e3s: here too

PR #170, opened 2016-11-17
Remove commented restrictions because they are still in ReDoc

XaosSintez:


2e3s:
does this solve the problem for you locally?

2e3s:
well, it doesn't for me. Somehow it gets kept in `swagger.yaml`

XaosSintez:
@2e3s we need to wait for tests. Then, preview will be available

XaosSintez:
@2e3s 
http://rebilly.github.io/RebillyAPI/preview/remove-commented-restrictions/

They are removed now, check please

PR #173, opened 2016-11-23
Add sortable fields example

XaosSintez:
Preview: http://rebilly.github.io/RebillyAPI/preview/sort-example/

Waiting for @RomanGotsiy thoughts about this. May be we can make it better, like `allOf` for parameters

### DIFF
--- a/spec/paths/customers.yaml
+++ b/spec/paths/customers.yaml
@@ -11,14 +11,14 @@ get:
     - $ref: "#/parameters/collectionQuery"
     - $ref: "#/parameters/collectionExpand"
     - $ref: "#/parameters/collectionFields"
-    - x-sortableEnum:
-      - $ref: "#/parameters/collectionSort"
-      - items:
-          enum:
-            - id
-            - email
-            - createdTime
-            - updatedTime
+    - x-rebillyMerge:
+        - $ref: "#/parameters/collectionSort"
+        - items:
+            x-sortableEnum:
+              - id
+              - email
+              - createdTime
+              - updatedTime
   responses:
     200:
       description: A list of Customers was retrieved successfully
--- a/spec/paths/customers.yaml
+++ b/spec/paths/customers.yaml
@@ -11,14 +11,18 @@ get:
     - $ref: "#/parameters/collectionQuery"
     - $ref: "#/parameters/collectionExpand"
     - $ref: "#/parameters/collectionFields"
-    x-rebillyMerge:
+    - x-rebillyMerge:
       - $ref: "#/parameters/collectionSort"
       - items:
          enum:
-           - id
-           - email
-           - createdTime
-           - updatedTime
+           - "id"
+           - "-id"
+           - "email"
+           - "-email"
+           - "createdTime"
+           - "-createdTime"
+           - "updatedTime"
+           - "-updatedTime"
 
   responses:
     200:
--- a/spec/paths/customers.yaml
+++ b/spec/paths/customers.yaml
@@ -11,7 +11,7 @@ get:
     - $ref: "#/parameters/collectionQuery"
     - $ref: "#/parameters/collectionExpand"
     - $ref: "#/parameters/collectionFields"
-    - x-rebillyMerge:
+    - x-sortableEnum:
       - $ref: "#/parameters/collectionSort"
       - items:
           enum:
--- a/spec/paths/customers.yaml
+++ b/spec/paths/customers.yaml
@@ -14,16 +14,16 @@ get:
     - x-rebillyMerge:
       - $ref: "#/parameters/collectionSort"
       - items:
-         enum:
-           - "id"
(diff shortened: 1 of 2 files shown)


spec/paths/customers.yaml
```diff
@@ -9,9 +9,17 @@ get:
...
-    - $ref: "#/parameters/collectionSort"
-    - $ref: "#/parameters/collectionFields"
     - $ref: "#/parameters/collectionExpand"
+    - $ref: "#/parameters/collectionFields"
+    - name: sort
+      in: query
+      type: string
+      description: The collection items sort field and order (prefix with "-" for descending sort).
+      enum:
+        - id
```
RomanGotsiy: From description seems that `"-id"`, `"-email"`, etc are valid values too. So you probably have to add these to enum if this spec will be using for validation in the future.

spec/paths/customers.yaml
```diff
@@ -9,9 +9,17 @@ get:
     - $ref: "#/parameters/collectionOffset"
     - $ref: "#/parameters/collectionFilter"
     - $ref: "#/parameters/collectionQuery"
-    - $ref: "#/parameters/collectionSort"
-    - $ref: "#/parameters/collectionFields"
     - $ref: "#/parameters/collectionExpand"
+    - $ref: "#/parameters/collectionFields"
+    - name: sort
```
slavcodev: Wonder if we can use `allOf` here

XaosSintez: Yes, that's why i asked Roman's help, because there's no allOf as i see for params

RomanGotsiy:
@XaosSintez enum is a correct way to do it. I understand it requires a lot of duplication. But unfortunately, `allOf` doesn't work in parameters.
What if I add additional bundle step which will build these enums automatically?

XaosSintez:
Yes, it will be better. Can you please show example, how it will look like

spec/paths/customers.yaml
XaosSintez: This is more for our frontend guys. Because it may be not one field, like
`?sort=-createdTime,id`

RomanGotsiy:
@XaosSintez I'm not sure right now about the exact syntax.
I will show it tomorrow morning, OK?

spec/paths/customers.yaml
Teknologica: So multiple sorting would be supported in the future?

XaosSintez: @Teknologica actually it is already supported with format i gave above :-)

RomanGotsiy:
OK,

First. As I have understood this param can accept a array of comma-separated values. So the proposed description is technically not correct.

The correct looks like the following:
```yaml
type: array
collectionFormat: csv
description: The collection items sort field and order (prefix with "-" for descending sort).
items:
  type: string
  enum:
    - id
    - -id
    - email
    - -email
    - ...
```

BUT this is not currently supported by ReDoc at the moment :( But this is simple thing so I can implement this during weekends.

Next. To omit duplication I can implement a build step which will inline param and merge. The proposed syntax:

```yaml
- ....
- $ref: "#/parameters/collectionExpand"
- x-merge:
  - $ref: "#/parameters/collectionSort"
  - items:
     enum:
      - id
      - -id
      - …
```
After processing this will be 

Why to use `x-merge` and not `allOf`? Because `allOf` has not obvious semantic and may be hard to implement while just merge is pretty simple.

If you all are OK with it, I can implement this during weekends

RomanGotsiy:
@XaosSintez I have just pushed a commit with corresponding post-processing plugin to master (planed to make this as PR but just forgot to switch branch :))

So rebase this branch on master and you'll be able to use it.

The syntax is:

```yaml
x-rebillyMerge:
  - $ref: "#/parameters/collectionSort"
  - items:
     enum:
      - aaa
      - bbb
```

slavcodev:
Wonder, why do we need to specify attributes twice, for `asc` and `desc` order, like `id` and `-id`? I think all attributes which we accept sorting, support the both order direction.

RomanGotsiy:
@slavcodev because without it if you use swagger-based validation/testing in the future validation for `"-id"` will fail.

Also, once ReDoc has API console it won't allow user to put values with `-`

slavcodev:
Okay, but could your extension `x-merge` to do this job?

RomanGotsiy:
I believe we should use another keyword for this. `x-rebillyMerge` can be used in other places too.
Something like `x-sortableEnum` or maybe you have better suggestions?

slavcodev:
`x-sortableEnum` looks okay, may be `x-sortable-enum`. So can we enumerate only unique attribute names, and the extension which prepare `swagger.json` will duplicate with `-`?

RomanGotsiy:
yes, sure.
I will implement this plugin tomorrow morning.

PR #174, opened 2016-11-24
Move security defs to correct place

RomanGotsiy:


RomanGotsiy:
@XaosSintez review plz
You added `SDK samples` section but left security definitions tag in the end and it should be in the end of security definitions section

PR #181, opened 2016-12-06
Fixes missing required fields for Disputes (GWY-5017)

dkotin:
Resolves: GWY-5017


XaosSintez:
You should also add them to `required` section in backend

dkotin:
only one was really needed - the rest are checked by type and relations. I've carefully checked.

PR #188, opened 2016-12-07
Add payment methods

colinbird:


spec/definitions/PaymentInstrument.yaml
```diff
@@ -4,10 +4,4 @@ required:
...
 properties:
   method:
-    description: The payment method
-    type: string
-    enum:
-      - ach
-      - cash
-      - payment-card
-      - paypal
+    $ref: "#/definitions/Method"
```
colinbird: is this ok, or breaks the discriminator ?

spec/definitions/Method.yaml
```diff
@@ -6,4 +6,5 @@ enum:
   - china-unionpay
   - payment-card
   - paypal
-
+  - OchaPay
```
slavcodev: Just I not sure there we need to use brand name.

adamaltman: Yes, and I think we use it, as well as `China UnionPay`.

slavcodev: Usage of brand name as gateway name makes sense for me, but there is instrument type, it's seems our internal things. It's looks confusing to have `payment-card` and `OchaPay` in one enum.

adamaltman: It seems we made a policy to use brand names with appropriate brand casing where brands are used.  For example, above we have `paypal`, but the brand is actually `PayPal`.

PR #194, opened 2016-12-13
Add point to activate user account by token (GWY-5088)

maxyengine:
Resolves: GWY-5088


spec/paths/activation@{token}.yaml
```diff
@@ -0,0 +1,26 @@
+parameters:
+  - name: token
+    in: path
+    description: The token string
+    type: string
+    required: true
+post:
```
XaosSintez: As i understand, this endpoint does not require any user auth. So, it should be described, like here:
https://github.com/Rebilly/RebillyAPI/blob/0f1def37a88c5bc53c7c47ef3f1898b5db3b9157/spec/paths/tokens@%7Btoken%7D.yaml#L14-L14

PR #201, opened 2016-12-23
Add minimum-order-amount & total-redemptions

2e3s:


2e3s:
@RomanGotsiy I'm having a problem. `RedemptionRestriction` has fewer points than `CouponRestriction` does. I've created `RedemptionRestriction`, and specified there the same points which have the same names as in `CouponRestriction`. But there is an empty list in UI. I suspect that this is because they all have `$ref: "#/definitions/CouponRestriction"`. I'm not sure how to deal with it.
So the task is to use different `RedemptionRestriction` and `CouponRestriction` with the same restrictions like `minimum-order-amount` etc.

RomanGotsiy:
@2e3s I have a solution but it requires small updates in ReDoc. I will let you know once I'm done.

RomanGotsiy:
@2e3s the solution:

create generic restriction type:
```yaml
description: All restriction
type: object
discriminator: "type"
properties:
  type:
    description: Restriction type
    type: string
```

and all of the restrictions should be inherited from it, e.g.:

```yaml
description: minimum-order-amount restrictions
allOf:
  -
    $ref: "#/definitions/GenericRestriction"
  -
    type: "object"
    ...
```

Then inherit `CouponRestriction` and `RedemptionRestriction` from `GenericRestriction` and limit possible `type` values with `enum`:

```yaml
description: Redemption restrictions
type: object
allOf:
  -
    $ref: "#/definitions/GenericRestriction"
  -
    type: "object"
    properties:
      type:
        description: Redemption's additional restriction type
        type: string
        enum:
          - discounts-per-redemption
          - restrict-to-invoices
          - restrict-to-plans
          - restrict-to-subscriptions
          - minimum-order-amount
```

I have all the changes done locally so if you don't mind I will push it to this branch

2e3s:
@RomanGotsiy please, do it.

RomanGotsiy:
@2e3s pushed.
http://rebilly.github.io/RebillyAPI/preview/gwy-5156-coupon-restrictions/#tag/Coupons

Seems to work as intneded

PR #232, opened 2017-01-09
Lists

XaosSintez:


spec/swagger.yaml
```diff
@@ -163,6 +163,11 @@ tags:
       Layouts are used to hold collections of plans. A layout can be used to
       power a pricing page. You can make multiple layouts, and use rules to
       target them to different audiences.
+  - name: Lists
+    description: |
+      Lists may be used in Rules conditions and can be edited without edition Rule itself.
```
adamaltman:
```
Lists contain sets of values and may be referenced within Rules criteria.  You may grant permissions to edit Lists to different people than those who can edit Rules.  It may be useful if your workflow involves frequent updates to value sets used in criteria.
```

spec/definitions/List.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - $ref: "#/definitions/ResourceId"
+  version:
+    description: List version
+    type: integer
+    readOnly: true
+    minimum: 1
+  name:
+    description: List name
+    type: string
+  path:
```
adamaltman: I think we can remove this member, `path`.

spec/definitions/List.yaml
```diff
@@ -0,0 +1,44 @@
...
+    description: Rules path to which List can be applied
+    type: string
+  values:
+    description: List values
+    type: array
+    items:
+      allOf:
+        - $ref: "#/definitions/ResourceId"
```
adamaltman: This is not good IMO, it may be a collection of resource ids, but it may be a collection of other types of values.  It may be quite similar to the `ResourceId` definition, but I'm not sure the description makes sense:

> The resource ID.  Defaults to UUID v4.

spec/swagger.yaml
```diff
@@ -163,6 +163,9 @@ tags:
       Layouts are used to hold collections of plans. A layout can be used to
       power a pricing page. You can make multiple layouts, and use rules to
       target them to different audiences.
+  - name: Lists
+    description: |
+      Lists contain sets of values and may be referenced within Rules criteria.  You may grant permissions to edit Lists to different people than those who can edit Rules. It may be useful if your workflow involves frequent updates to value sets used in criteria.
```
adamaltman: I think you should wrap this into multiple lines here for readability of the spec.

Teknologica:
I wonder if adding a `description` alongside the the `name` would be useful for merchants. Some lists could have similar names and trying to discern what makes them different by looking at the count of values, or the values themselves, could make it complicated for them.

adamaltman:
I think we should try without it first.

PR #242, opened 2017-01-30
Remove Histograms specs out of main specs (GWY-5305)

wmdmgv:
Ticket https://jira.ilabsinc.com/browse/GWY-5305


2e3s:
`HistogramData` should also be removed.

PR #244, opened 2017-01-31
Add billingAddressId  to Transaction specs (GWY-5395)

wmdmgv:


spec/definitions/Transaction.yaml
```diff
@@ -161,6 +161,10 @@ properties:
     format: integer
   paymentInstrument:
     $ref: "#/definitions/PaymentInstrument"
+  billingAddressId:
```
adamaltman: This is wrong.  Look at the rest of the api, we use the term `contact` instead of `address`.

spec/definitions/Transaction.yaml
```diff
@@ -161,6 +161,10 @@ properties:
...
+    description: Billing Address ID
+    type: number
```
vova07: `billingContactId` will be an external `id` so it will be a string instead.

PR #247, opened 2017-02-01
Set format: bin to support bin range search

colinbird:


spec/definitions/PaymentCard.yaml
```diff
@@ -15,6 +15,7 @@ properties:
   bin:
     description: The card's bin (the PAN's first 6 digits)
     type: string
+    format: bin
```
slavcodev: Hmm, I can't find the information about this format in the `OpenApi` docs, in the `JSON-Schema` either.
could you please post proof-link?

colinbird: the frontend team suggested this format in order to treat bins specifically

PR #250, opened 2017-02-06
Adjust Product definitions

viceromag:


spec/definitions/Product.yaml
```diff
@@ -44,9 +48,14 @@ properties:
...
       - 30070
-  customFields:
-    description: Any custom object that may be stored with product (see custom fields)
+  requiresShipping:
+    description: If the product requires shipping, shipping calculations will be applied
+    type: boolean
+  accountingCode:
+    description: The product accounting code
     type: string
+  customFields:
```
2e3s: So you need to change https://github.com/Rebilly/RebillyAPI/blob/master/spec/paths/custom-fields%40%7Bresource%7D%40%7Bname%7D.yaml and https://github.com/Rebilly/RebillyAPI/blob/master/spec/paths/custom-fields%40%7Bresource%7D.yaml

PR #256, opened 2017-02-15
Remove deprecated Transaction.paymentCardId field

slavcodev:


spec/definitions/Transaction.yaml
```diff
@@ -129,6 +129,14 @@ properties:
           message:
             description: The raw response message
             type: string
+  gatewayName:
+    description: Payment Gateway name, available only when method is 'payment_card', else null
```
adamaltman: This seems wrong.  It should be available for some other methods too.

slavcodev: Yes, I've copy-pasted and did not notice the erroneous description. Will fix soon

PR #257, opened 2017-02-16
Add bin/last4 to customer resource

dara123:


spec/definitions/Customer.yaml
```diff
@@ -16,6 +16,12 @@ properties:
   lastName:
     description: The customer last name
     type: string
+  bin:
```
slavcodev: I believe it's wrong to put that in the root, I think it should be in the `defaultPaymentInstrument` and only for payment cards.

PR #258, opened 2017-02-17
Add 3DS data to Transaction

slavcodev:


spec/definitions/Transactions/ThreeDSecureResult.yaml
```diff
@@ -0,0 +1,29 @@
...
+  enrolled:
+    description: Is the cardholder enrolled in 3DSecure
+    type: string
+    enum:
+     - "yes"
+     - "no"
+     - "invalid card/timeout"
+     - "unavailable"
+  authenticated:
+    description: The ThreeDSecure entry Auth Response Status
```
adamaltman: I think change `ThreeDSecure` to `3D Secure` in the description.

PR #259, opened 2017-02-17
Implement global webhooks API spec (GWY-5484)

wmdmgv:
We should implement the Webhook SPEC.
I think that we need next actions:
Retrieve webhook/s
Create/Update webhook
Delete webhook
Trigger test webhook


spec/definitions/GlobalWebhook.yaml
```diff
@@ -0,0 +1,18 @@
...
+properties:
+  id:
+    description: The webhook identifier string
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  eventType:
+    description: System event type
+    type: string
+  url:
+    description: URL which will be callen when the given event happens
```
adamaltman: URL that will be requested when the given event occurs.

spec/definitions/GlobalWebhook.yaml
```diff
@@ -0,0 +1,35 @@
...
+      $ref: "#/definitions/GlobalWebhookEventType"
+  method:
+    type: string
+    enum:
+      - GET
+      - POST
+      - PUT
+      - PATCH
+      - DELETE
+  url:
```
slavcodev: `format: uri`

spec/definitions/GlobalWebhook.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - method
+  - url
+  - credentialHash
+properties:
+  id:
+    description: The webhook identifier string
+    readOnly: true
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  eventType:
```
slavcodev: How to add hook for all events?

spec/paths/webhooks@{id}.yaml
```diff
@@ -0,0 +1,55 @@
+parameters:
```
slavcodev: Missing `DELETE` hook, isn't it?

spec/paths/webhooks@{id}@disable.yaml
```diff
@@ -0,0 +1,24 @@
+post:
```
slavcodev: I believe it useless endpoint, it is more useful to add property to the webhook resource IMO

spec/paths/webhooks@{id}@trigger.yaml
```diff
@@ -0,0 +1,26 @@
+post:
```
slavcodev: what is the payload of this request? what payload will be used for webhook?

spec/swagger.yaml
```diff
@@ -416,6 +421,12 @@ parameters:
       - application/json
     default: application/json
     description: The response media type
+  eventType:
```
slavcodev: we already have `systemEventType`, can we use that parameter?

spec/definitions/GlobalWebhook.yaml
wmdmgv: for all events - empty array

spec/paths/webhooks@{id}.yaml
wmdmgv: we have `disable` only, not `delete`

spec/swagger.yaml
wmdmgv: we have discussed it with Andrei and decided that this enum do not fit for webhook correctly and a new list require.

spec/definitions/GlobalWebhook.yaml
slavcodev: Then `eventType` make no sense and confusing. Would be better smth like `eventsFilter`, `eventsCriteria`

spec/paths/webhooks@{id}.yaml
slavcodev: Is it requirements? What is the reason to disallow removing?

spec/definitions/GlobalWebhook.yaml
vova07: I propose to have `events` array of string or an empty one.

spec/paths/webhooks@{id}.yaml
vova07: We can do soft delete only, so I think is better to disable the subscribers instead of deleting them.

spec/paths/webhooks@{id}@disable.yaml
vova07: We can disable web hook only one time.
Also it can be done by mistake in case of a body parameter.
I opt for separate URI in this case.

spec/paths/webhooks@{id}@trigger.yaml
vova07: I think it will be an empty payload. Like we have in `Rules test Triggering`.
Or we can create a fake event and fake payload, to use instead.

spec/definitions/GlobalWebhook.yaml
slavcodev: well, empty `events` or `eventType` list means none of events will trigger hook, not all events, so these names semantically invalid, IMO

spec/paths/webhooks@{id}.yaml
slavcodev: Could you @vova07 provide more information, why we can do only `soft` delete, and what is the meaning from customer side (not technical, I know we can remove record from DB because it has relationship)?

spec/paths/webhooks@{id}@disable.yaml
slavcodev: I'm not sure I understand you. One time? Do you mean merchant can't enable it back? What about hooks collection, the disabled hooks will be the collection response? How merchant will treats so records?

I think I disagree with such design. For me
- Operation `disable` means the entity property, and means ability to enable back
- Operation `delete` (no matter soft of hard on backend) means removing from merchant list and make unusable (soft-removing case)

so let's clarify requirements, what do we want to implement?

spec/definitions/GlobalWebhook.yaml
vova07: Than I think we should go like Stripe with 2 attributes:
`allEvents => true and events => []`

This is what Stripe return: `"enable_all_events":true,"enabled_events":[]`

spec/paths/webhooks@{id}.yaml
vova07: Yes, we can do soft delate only because of relations and because of Tracking.
Non technical - no reason
But as I said, I'm worried about probably mistakes, and probably we should use `DELETE /webhooks/{id}` as we discuss with Gleb yesterday instead.

spec/paths/webhooks@{id}@disable.yaml
vova07: Comment above clarify it. And yes, I think we should go with `DELETE` from now.

spec/definitions/GlobalWebhook.yaml
```diff
@@ -0,0 +1,46 @@
+type: object
+required:
+  - eventsFilter
```
vova07: It's not required. By default is an empty array.

spec/definitions/GlobalWebhook.yaml
```diff
@@ -0,0 +1,46 @@
...
+  method:
+    type: string
+    enum:
+      - GET
+      - POST
+      - PUT
+      - PATCH
+      - DELETE
+  url:
+    description: URL that will be requested when the given event occurs.
```
vova07: `URL that will be triggered when the given event occurs.` ?

spec/swagger.yaml
```diff
@@ -274,6 +274,11 @@ tags:
       website is related to each invoice and each payment gateway account.  This feature
       would allow you to have gateway accounts that are related to multiple websites, or
       exclusive to particular websites.  And gives you more control over your business.
+  - name: Webhooks
+    description: |
+      Webhooks are designed to notify your systems when certain/all registered events happen in near real-time.
+      They allow you to collect information about those events. Rebilly can send these information via
+      POST to a URL of your choice.
```
vova07: `POST to an URL`

spec/definitions/GlobalWebhook.yaml
```diff
@@ -0,0 +1,45 @@
+type: object
+required:
+  - method
+  - url
+  - credentialHash
+properties:
```
slavcodev: wonder if something like `title` or `description` would be useful for webhook?

spec/paths/webhooks@trigger@preview.yaml
```diff
@@ -0,0 +1,29 @@
+post:
```
slavcodev: I believe this URI conflicts with `webhooks/{id}` and is confusing

spec/definitions/GlobalWebhook.yaml
vova07: I don't think that those attributes are useful here.

spec/paths/webhooks@trigger@preview.yaml
vova07: Slava why it should conflict ? They are different URI levels, not sure you are right here.
Of course in case we had something like `webhooks/{id}/some-action` it can conflict but we don't have any other URLs, only those 2.

spec/definitions/GlobalWebhook.yaml
slavcodev: okay, if you are sure, I just wonder how merchant will distinguish webhooks in the grid, in case he will have a couple of them.

spec/paths/webhooks@trigger@preview.yaml
slavcodev: `webhooks/trigger/preview` is a sub-path of `webhooks/trigger`, isn't it? the second conflicts `webhooks/{id}`.

I believe the URL is not one string, it is multi-level structure of sections, is a chain, and each level has a meaning.

Not sure I understand what do you and Gleb mean.

spec/definitions/GlobalWebhook.yaml
vova07: If you think that this can be useful, and you are disagree with me, we should discuss this subject in our team room, because I can't give you here pro and cons.

spec/paths/webhooks@trigger@preview.yaml
vova07: I understand your opinion, but at the same time, we should have the same `root`, because all this is related with webhooks.
If you think it's really bad to have such URLs I think we can go with `preview` root, and use that for all our context in case we need to preview something.

slavcodev: I believe is wrong and really confusing, I'm doubt it RESTful.

vova07: You doubt about `/webhooks/trigger/preview` or about `/preview/trigger/global-webhook/` ?

slavcodev: I'm doubt about proposed URI

PR #274, opened 2017-03-06
Discounts in Invoice refactor

XaosSintez:
1. We have all Coupons applied info in Invoice. We need them to represent all Coupons applied in Invoice View in FE. Each of them has redemption code, description (to show in invoice  something like `Discount for all new customers $10` )  and total discount amount for each coupon applied to that invoice
2. We have calculated `discountAmount` in each Item that we need for tax calculations

### DIFF
--- a/spec/definitions/Coupon/InvoiceDiscount.yaml
+++ b/spec/definitions/Coupon/InvoiceDiscount.yaml
@@ -10,7 +10,7 @@ properties:
     allOf:
       - $ref: "#/definitions/ResourceId"
   amount:
-    description: Discount amount
+    description: Total amount that was discounted by this Coupon
     type: number
     format: double
   description:
--- a/spec/definitions/Coupon/InvoiceDiscount.yaml
+++ b/spec/definitions/Coupon/InvoiceDiscount.yaml
@@ -13,7 +13,3 @@ properties:
     description: Discount amount
     type: number
     format: double
-  relatedItemId:
-    description: Related item's ID (can be null if Discount was applied to whole Invoice)
-    allOf:
-      - $ref: "#/definitions/ResourceId"
--- a/spec/definitions/Coupon/InvoiceDiscount.yaml
+++ b/spec/definitions/Coupon/InvoiceDiscount.yaml
@@ -1,0 +1,19 @@
+type: object
+readOnly: true
+properties:
+  redemptionCode:
+    description: Coupon's redemption code
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  redemptionId:
+    description: Redemption ID
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  amount:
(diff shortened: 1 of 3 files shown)


spec/definitions/Invoice.yaml
```diff
@@ -47,6 +47,12 @@ properties:
     readOnly: true
     items:
       type: string
+  discounts:
```
2e3s: Do we need it here? What for? It's kind of a duplication of `InvoiceItem`.

spec/definitions/Coupon/InvoiceDiscount.yaml
```diff
@@ -0,0 +1,18 @@
...
+properties:
+  redemptionCode:
+    description: Coupon's redemption code
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  redemptionId:
+    description: Redemption ID
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  amount:
```
2e3s: What does this `amount` mean?

XaosSintez: This is amount that was discounted by this Coupon in total.
For example if it is a percent type Coupon, it can discount $1.5 in total from 3 items. Let me know if i answered not clearly

XaosSintez: May be we should rename it to `discountAmount`

2e3s: No, I think you shouldn't, this name doesn't make any difference. Probably you just to clarify this in the description.

spec/definitions/Coupon/InvoiceDiscount.yaml
```diff
@@ -0,0 +1,18 @@
+type: object
+readOnly: true
+properties:
```
2e3s: How would you know which discount relates to which items? And is this information needed at all?

2e3s: What about per-invoice discounts? Will they be just prorated by design?

XaosSintez: After some discussions we decided that this info in not really necessary. Discount amount for each item is required for tax calculations, but no reason to have info about each discount for each item, discount total for each item is enough imho.

But if you can suggest use case where it is necessary - let me know

2e3s: Well, I don't think that we have to bind solely to taxes.. but neither can I think of any use of this information.

What about per-invoice discounts? Will they be just prorated by design for all items?

XaosSintez: > Will they be just prorated by design for all items?
Yeah. For all debit items for now. 
Later we may add `discountable` property. May be.

PR #277, opened 2017-03-09
Add preferences to User resource

XaosSintez:


spec/definitions/User.yaml
```diff
@@ -78,6 +78,7 @@ properties:
     type: string
   preferences:
     description: User preferences like timezone, language and many more. This is an object with custom properties.
+    type: object
```
slavcodev: Hmm, is not that preference pre-defined, is it?

XaosSintez: They are not pre defined

slavcodev: Wonder how they can be variable if client (at least our UI) requires specific ones?

XaosSintez: They don't require it. They set it and use it. Actually, backend has no idea, which properties will be there and backend doesn't use it at all.
It is totally defined by FE and used only by FE

slavcodev: Alright.

PR #282, opened 2017-03-14
Fix Dispute reason code's

wmdmgv:
Fix format of enum reason code's.  Because we have a string type in api json and db, but in api spec REQUEST SAMPLES :
`"reasonCode": 1000, `

and add few new reason code's .


spec/definitions/Dispute.yaml
```diff
@@ -33,122 +33,135 @@ properties:
...
+      - "UA18"
+      - "UA20"
+      - "UA21"
+      - "UA22"
+      - "UA23"
+      - "UA28"
+      - "UA30"
+      - "UA31"
+      - "UA38"
+      - "duplicate"
```
XaosSintez: Are you sure about this values? Looks like this is not the reasonCode, but it's description

wmdmgv: @XaosSintez we had to change these reason code's.

PR #292, opened 2017-03-24
Rename GetGateway to GET

dara123:


Zaphell:
https://github.com/Rebilly/RebillyAPI/blob/c6ee884cd4f20d6a5ea9d6b221b73170e052c736/spec/definitions/GatewayAccountConfig/GET.yaml#L15

Should be ```GET gateway account ID"

PR #295, opened 2017-03-30
Adjust custom events

creocoder:


spec/definitions/CustomEvents/CustomEvent.yaml
```diff
@@ -20,7 +20,7 @@ properties:
       - before
       - after
   scheduleInstruction:
-    $ref: "#/definitions/ScheduleMethod"
+    $ref: "#/definitions/CustomEventScheduleInstruction"
```
slavcodev: What is the reason of this change? Why the exiting one can't be used?

creocoder: @slavcodev Not all values are supported.

PR #304, opened 2017-04-19
Better custom events descriptions

creocoder:


spec/definitions/Scheduling/Methods/date-interval.yaml
```diff
@@ -5,6 +5,7 @@ allOf:
     properties:
       duration:
         type: integer
+        description: The duration of the unit
```
slavcodev: Actually it is not a duration of the one unit, it is number of units.

PR #315, opened 2017-05-22
Invoice status

XaosSintez:


spec/definitions/Invoice.yaml
```diff
@@ -66,6 +66,25 @@ properties:
...
+    enum:
+     - "draft"
+     - "issued"
+     - "past-due"
+     - "paid"
+     - "abandoned"
+     - "voided"
+  delinquentCollectionPeriod:
+    type: integer
+    description: Days count when Invoice was past due.
```
dkotin: Sounds a bit strange, idk, need to rephrase somehow...
```Days count from past due date```  mb?

File: spec/definitions/Invoice.yaml [hunk 2]
```diff
@@ -66,6 +66,25 @@ properties:
...
+    readOnly: true
+  collectionPeriod:
+    type: integer
+    description: Days count when Invoice was issued.
```
dkotin: Should sound more like days count from the moment Invoice has been issued, right? May be something shorter, like
```Days count from issued date```?

PR #320, opened 2017-05-30
Add transaction cancel path and deprecate payment cancel

slavcodev:


spec/paths/transactions@{id}@cancel.yaml
```diff
@@ -2,10 +2,9 @@ parameters:
   - $ref: "#/parameters/resourceId"
 post:
   tags:
-    - Payments
-  summary: Cancel a scheduled payment
-  description: |
-    Cancel a scheduled payment with specified identifier string
+    - Transactions
+  summary: Cancel a pending or suspended transaction
+  description: Cancel a scheduled payment but not handled transacitons
```
adamaltman: typo here in `transactions`.  Also shouldn't it read:

> Cancel a scheduled transaction.  Once handled a transaction cannot be canceled.

PR #325, opened 2017-06-06
Add bulk-data export first endpoint

colinbird:


dara123:
Is it only for transactions?

colinbird:
@dara123 to start yes, the requirement is just for transactions.

spec/paths/export@transactions.yaml
```diff
@@ -0,0 +1,39 @@
+post:
```
dara123: should it be GET?

colinbird: no, you can see some examples here 

https://docs.google.com/document/d/1Ftx4RUnOTwi2grkK6GLjs0_q53tlDP-w5FubaH6Fm4k/edit

spec/paths/export@transactions.yaml
```diff
@@ -0,0 +1,69 @@
...
+  tags:
+    - Exports
+  summary: Request a bulk export of Transactions
+  description: |
+    Request a bulk export of Transactions
+  parameters:
+    - $ref: "#/parameters/collectionFilter"
+    - $ref: "#/parameters/collectionSort"
```
colinbird: going to put the filter and sort in the body, to match the POST call

spec/paths/export@transactions.yaml
colinbird: GET added

spec/definitions/Export.yaml
```diff
@@ -0,0 +1,43 @@
+type: object
+required:
+  - id
```
slavcodev: `required` section in specs is for properties which are supposed to be input by client, thus `id`, `status`, `ceratedTime` should not be here.

spec/definitions/Export.yaml
```diff
@@ -0,0 +1,43 @@
...
+    type: string
+    enum:
+      - "pending"
+      - "completed"
+  format:
+    description: Format
+    type: string
+    enum:
+      - "csv"
+  fileUri:
```
slavcodev: `readOnly` right?

spec/definitions/Export.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - $ref: "#/definitions/ServerTimestamp"
+  completedTime:
+    description: Export request completed time
+    allOf:
+      - $ref: "#/definitions/ServerTimestamp"
+  _links:
+    type: array
+    description: The links related to resource
+    readOnly: true
+    minItems: 1
```
slavcodev: please update the min and max number, I see 2 links specified below.

spec/paths/export@transactions.yaml
slavcodev: @dara123 the `GET` is idempotent and safe, it's definitely inappropriate to create the export task.

spec/paths/export@transactions.yaml
```diff
@@ -0,0 +1,69 @@
+post:
```
slavcodev: I wonder, would be `/exports/transactions` more semantically better?

spec/paths/export@transactions.yaml
```diff
@@ -0,0 +1,69 @@
...
+        type: object
+        required:
+          - format
+        properties:
+          format:
+            description: Export format
+            type: string
+            enum:
+              - csv
+  produces:
```
slavcodev: I think this can be omitted, since it is base for Rebilly API and specified in the `swagger.yaml` (same for `GET` operation).

spec/definitions/Export.yaml
```diff
@@ -21,8 +20,13 @@ properties:
     type: string
     enum:
       - "csv"
+  url:
```
slavcodev: that is not looks like url, I think would be better to find batter name, e.q. `exportUriPath` or something like this.

colinbird:
retest please

spec/definitions/Export.yaml
```diff
@@ -0,0 +1,47 @@
...
+  id:
+    description: The export identifier string
+    readOnly: true
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  status:
+    description: Status of export request
+    readOnly: true
+    type: string
+    enum:
```
XaosSintez: May be we can add `error` status just in case? Or it's not planned to get any error here? :-)

spec/definitions/Export.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - "pending"
+      - "completed"
+  format:
+    description: Format
```
XaosSintez: How about something like `Export resulting format` ?

spec/paths/exports@transactions.yaml
```diff
@@ -0,0 +1,55 @@
...
+      $ref: "#/responses/AccessForbidden"
+get:
+  tags:
+    - Exports
+  summary: Retrieve a list of export requests
+  description: |
+    Retrieve a list of export requests
+  responses:
+    200:
+      description: A list of transactions was retrieved successfully
```
XaosSintez: I think it's a copy-paste :-)

spec/paths/exports@transactions.yaml
```diff
@@ -0,0 +1,55 @@
...
+    Request a bulk export of Transactions
+  parameters:
+    - name: body
+      in: body
+      description: Export Request
+      required: true
+      schema:
+        $ref: "#/definitions/Export"
+  responses:
+    201:
```
XaosSintez: It's `201`, not `202 Accepted`, right?

spec/definitions/Export.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - "completed"
+      - "error"
+  format:
+    description: Export result format
+    type: string
+    enum:
+      - "csv"
+  exportUriPath:
+    description: URI to export
+    example: /v2.1/transactions?sort=createdTime
```
2e3s: 1) Does it make any sense to specify `/v2.1/transactions?` here? 
2) Maybe make 3 explicit fields: `sort`, `criteria`, `filter`?
2.1) Probably as subobjects of `arguments`

spec/definitions/Export.yaml
slavcodev: What errors do you mean? If backend will hit errors, we will do the best to finish the export.

spec/paths/exports@transactions.yaml
slavcodev: Yes, `201`.

spec/definitions/Export.yaml
colinbird:
```
{
    "format": "csv",
    "arguments": {
        "sort": "",
        "filter": "",
        "criteria": ""
    }
}
```

colinbird: is this what you suggest @2e3s  ?

2e3s: yes 👍

spec/definitions/Export.yaml
```diff
@@ -0,0 +1,61 @@
...
+        description: The collection items filter requires a special format. Use "," for multiple allowed values. Use ";" for multiple fields. See the filter guide for more options and examples about this format.
+        type: string
+      sort:
+        type: array
+        description: The collection items sort field and order (prefix with "-" for descending sort).
+        items:
+          type: string
+      q:
+        type: string
+        description: The partial search of the text fields.
```
2e3s: not partial.. more like fulltext

colinbird: thanks

dara123:
Move it to experimental API: https://github.com/Rebilly/RebillyReportsAPI/pull/18

PR #329, opened 2017-06-15
Add Risk Metadata API

XaosSintez:


spec/definitions/RiskMetadata.yaml
```diff
@@ -0,0 +1,25 @@
+type: object
+properties:
+  ipAddress:
```
slavcodev: I think we can avoid this level

XaosSintez: It's the grouping we discussed before. On same level will be `fingerprint`, `httpHeaders` and so on

slavcodev: in current specs, let say the Transaction JSON will show like

~~~json
{
  "riskMetadata": {
    "ipAddress": {
      "ipAddress": "192.168.1.1",
      "isVPN": true
    },
    "fingerprint": "foo"
  }
}
~~~

I think would be better to modify to

~~~json
{
  "riskMetadata": {
    "ipAddress": "192.168.1.1",
    "isVPN": true,
    "fingerprint": "foo"
  }
}
~~~

PR #334, opened 2017-07-10
Add refund link to Transaction

slavcodev:


spec/definitions/Transactions/Transaction.yaml
```diff
@@ -271,3 +271,4 @@ properties:
       - $ref: "#/definitions/LeadSourceLink"
       - $ref: "#/definitions/ApprovalUrlLink"
       - $ref: "#/definitions/CancelUrlLink"
+      - $ref: "#/definitions/RefundUrlLink"
```
XaosSintez: I wonder if you should update `maxItems`

slavcodev: Done, thank you

PR #336, opened 2017-07-12
Add Gpaysafe and Jeton to spec

dara123:


spec/definitions/Gateways/AcquirerName.yaml
```diff
@@ -29,10 +29,12 @@ enum:
   - FundSend
   - GlobalCollect
   - Global East
+  - Gpaysafe
   - Heartland
   - HSBC
   - iCheque
   - Ilixium
+  - Jetton
```
colinbird: typo here

PR #337, opened 2017-07-14
Add update-password endpoint to Profile API

XaosSintez:


spec/paths/profile@update-password.yaml
```diff
@@ -0,0 +1,31 @@
...
+    Updates user's password with the specified newPassword. And checks if currentPassword matches the actual one.
+  parameters:
+    - name: body
+      in: body
+      description: currentPassword and newPassword
+      required: true
+      schema:
+        $ref: "#/definitions/UpdatePassword"
+  responses:
+    201:
```
vova07: Do we create a new resource ?

XaosSintez: Well, we response with 201 when we cancel subscription or anything else like that. And also we send 201 in `users/{id}/password` too

vova07: hmmm ... seems incorrect.
Not sure how we miss that.

XaosSintez: We redirect to Profile after request (as we do  in many other places), so 201 seems correct

slavcodev: I think `201` is right. And yes we are created new resource - *new password`

slavcodev:
`profile@update-password` -> `profile@passwords` ?

XaosSintez:
@slavcodev there is no `passwords`, it is current user's password

PR #343, opened 2017-07-28
Fix multiple mistakes in spec

slavcodev:


spec/definitions/Method.yaml
```diff
@@ -2,15 +2,6 @@ type: string
...
-  - china-unionpay
-  - Flexepin
-  - Gpaysafe
-  - Jeton
   - payment-card
   - paypal
-  - OchaPay
-  - SMSVoucher
-  - UPayCard
-  - WeChat Pay
```
Zaphell: I think all those are considered payment methods...

slavcodev: Nope. They are not, they are wallet service name. Check the published docs, this are not in the drop down for payment method, its specs are not defined in https://github.com/Rebilly/RebillyAPI/tree/master/spec/definitions/PaymentInstuments

slavcodev:
This is invalid, the issue is deeper, we are messing high level categories like `payment-card`, `ach` and brand names like `AliPay`, `OchaPay`. Which seems to be different things. So need more investigation and another fix.

PR #349, opened 2017-08-07
Add isUsed flag to custom field resource

dara123:


spec/definitions/CustomField.yaml
```diff
@@ -31,6 +31,9 @@ properties:
   description:
     description: The custom field description
     type: string
+  isUsed:
+    description: The custom field flag whether it's being used.
```
Zaphell: A flag to represent that the custom field is used on a record of the resource.

PR #353, opened 2017-08-20
Add isDisputed flag in Transactions

viceromag:
Add isDisputed flag and links to Disputes to a Transactions response.

### DIFF
--- a/spec/definitions/Transactions/Transaction.yaml
+++ b/spec/definitions/Transactions/Transaction.yaml
@@ -266,7 +266,7 @@ properties:
     description: The links related to resource
     readOnly: true
     minItems: 1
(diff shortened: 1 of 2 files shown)


spec/definitions/Transactions/Transaction.yaml
```diff
@@ -276,3 +280,4 @@ properties:
       - $ref: "#/definitions/ApprovalUrlLink"
       - $ref: "#/definitions/CancelUrlLink"
       - $ref: "#/definitions/RefundUrlLink"
+      - $ref: "#/definitions/DisputeLink"
```
XaosSintez: You need to adjust `maxItems`

PR #354, opened 2017-08-23
Add File permalink

colinbird:


spec/definitions/File.yaml
```diff
@@ -63,3 +63,4 @@ properties:
     maxItems: 1
```
slavcodev: I need to adjust `maxItems`

colinbird: Should it be minItems 2 maxItems 2 ?

PR #355, opened 2017-08-25
Add Webhook Tracking

XaosSintez:


spec/definitions/Tracking/WebhookTracking.yaml
```diff
@@ -0,0 +1,39 @@
+type: object
+description: Webhook Tracking Requests.
+readOnly: true
+properties:
+  id:
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  eventName:
```
wmdmgv: And how can merchant to know which `method`, `url`, `headers`, and `query `  was called?

XaosSintez: I'm not sure about `headers`, can't they be kind of private? 
I will add all other fields except `query` - what do you mean by this?

wmdmgv: yeah, but AFAIR headers compose by merchant itself  that expected they need to their side (sent),... may be they need to see what they was received(i think, but maybe someone thinks differently), and `query `- is just query string of some parameters(if needed) to the  url. 
**Forexample** 
```
Method : GET 
Url : http://someurl.com/webbhook
query: company=rebilly
"headers":[{"name":"header-name","value":"header-value","active":true}]
```

XaosSintez: I think `query` is part of `url`. About `headers` - got it, thanks

wmdmgv: Okay, what about "red" (`continuous-integration/travis-ci`) checks below? hmmm

spec/definitions/Tracking/WebhookTracking.yaml
```diff
@@ -0,0 +1,53 @@
...
+    example:
+      "My-Header": "Cool-Value"
+  responseCode:
+    type: integer
+    description: HTTP code response
+  responseBody:
+    type: string
+    description: Response body received
+  payload:
+    type: string
```
2e3s: `object` maybe? to provide this as a convenient json

spec/definitions/Tracking/WebhookTracking.yaml
wmdmgv: The question about `eventName`.  We are stored   in `WebhookTrackingPayload.payload` this field : `"eventType"` , 
forexample :` "eventType" : "dispute-created"` may be we need to rename `eventName` to `eventType` ? or we will be translate  `"dispute-created"` to ` "Dispute created"` ? Or you did it like we have in website webhook tracking https://github.com/Rebilly/RebillyAPI/blob/4689c518594440ef6353c2e832325c4fdda94377/spec/definitions/Tracking/WebsiteWebhookTracking.yaml#L11  to consistent it in other places.?

spec/definitions/Tracking/WebhookTracking.yaml
```diff
@@ -0,0 +1,53 @@
...
+    description: Event name for which webhook was called
+  url:
+    type: string
+    description: Url where webhook was sent
+  method:
+    type: string
+    description: HTTP method which was used to send webhook
+    example: "POST"
+  headers:
```
wmdmgv: Why is it object? 
as i see `"headers":[{"name":"header-name","value":"header-value","active":true}]`
it's an array of objects.

spec/definitions/Tracking/WebhookTracking.yaml
```diff
@@ -0,0 +1,53 @@
...
+  createdTime:
+    description: The log created time
+    allOf:
+      - $ref: "#/definitions/ServerTimestamp"
+  _links:
+    type: array
+    description: The links related to resource
+    readOnly: true
+    minItems: 1
+    maxItems: 1
```
wmdmgv: `_links `is array with 1 item only?

spec/definitions/Tracking/WebhookTracking.yaml
XaosSintez: I don't insist on `eventName`, `eventType` looks ok too.

> or we will be translate "dispute-created" to "Dispute created"
Can you please explain?

spec/definitions/Tracking/WebhookTracking.yaml
XaosSintez: Yes, as we have it in many other places

spec/definitions/Tracking/WebhookTracking.yaml
XaosSintez: What do you mean? How we store it in backend?

wmdmgv: Forexample WebhookSubscriber.data :

```
{
	"credentialHash" : "2108c8b5-42dd-412d-9600-a95a2178d2ac",
	"headers" : [
		{
			"active" : true,
			"name" : "key1",
			"value" : "value1"
		},
		{
			"active" : true,
			"name" : "key2",
			"value" : "value2"
		}
	],
	"method" : "POST",
	"url" : "https://requestb.in/10z7qkt1"
}
```
and these headers will be send to merchant.

wmdmgv: As we discussed in PM, yeah,   you are right about object, becase sent only active headers. 
```
headers: {
  "My-Header": "My-Value",
  "My-Other-Header": "My-Other-Value",
}
```

spec/definitions/Tracking/WebhookTracking.yaml
wmdmgv: sorry  about 'translate', of course not, but i see https://github.com/Rebilly/RebillyAPI/blob/4689c518594440ef6353c2e832325c4fdda94377/spec/definitions/EventType.yaml#L5  
in eventType

PR #357, opened 2017-09-01
Add missing amount and gatewayAccountId

slavcodev:


spec/paths/paypal-accounts@{id}@activation.yaml
```diff
@@ -23,9 +23,17 @@ post:
...
             type: string
+          amount:
+            description: The amount to authorize
+            type: number
+            format: double
+            default: 1.00
           redirectURLs:
             description: Redirect URLs
             type: object
+          gatewayAccountIdId:
```
wmdmgv: Is it ok `IdId` double `id` ?

PR #362, opened 2017-09-19
Add better examples for Contacts in related resources

viceromag:
PHP SDK was updated to handle new way of adding Contacts to related resources, this PR adds up to date examples.


XaosSintez:
I think you can add more than one `email` or `phoneNumber` in examples. And one of them with `primary`

PR #367, opened 2017-10-09
Add attempt and source to Webhook Tracking

XaosSintez:


spec/definitions/Tracking/WebhookTracking.yaml
```diff
@@ -30,6 +30,15 @@ properties:
   payload:
     type: object
     description: Event's data information
+  source:
+    type: string
+    description: The Webhook's source
+    enum:
+     - "global-webhooks"
```
wmdmgv: Is we have `'global' `name for webhooks in API spec? 
there are doubts that merchants can correctly understand what this means.

spec/definitions/Tracking/WebhookTracking.yaml
```diff
@@ -30,6 +30,15 @@ properties:
...
+     - "rules-engine"
```
slavcodev: I think we should consider to use just `rules` or `rules-action`. Because right now we do not use in API the term `Rules Engine`

spec/paths/tracking@webhooks@{id}@attempts.yaml
```diff
@@ -0,0 +1,22 @@
+parameters:
+  - $ref: "#/parameters/resourceId"
+get:
+  tags:
+    - Tracking
+  summary: Retrieve chain for specified webhook
```
wmdmgv: In this case maybe will be better Retrieve chain of attempts for specified webhook

spec/paths/tracking@webhooks@{id}@history.yaml
```diff
@@ -0,0 +1,22 @@
+parameters:
+  - $ref: "#/parameters/resourceId"
+get:
+  tags:
+    - Tracking
+  summary: Retrieve related webhook attempts for specified webhook.
```
2e3s: "for the specified webhook tracking" I think

PR #368, opened 2017-10-09
Add Distance to RiskMetadata

viceromag:
Add distance to RiskMetadata.

### DIFF
--- a/spec/definitions/RiskMetadata.yaml
+++ b/spec/definitions/RiskMetadata.yaml
@@ -75,6 +75,6 @@ properties:
       "User-Agent": "Mozilla/5.0"
       "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"
   distance:
-    description: Distance between IP Address and Billing Address geolocations (kilometers)
+    description: Distance between IP Address and Billing Address geolocation (kilometers)
     type: integer
     readOnly: true
--- a/spec/definitions/RiskMetadata.yaml
+++ b/spec/definitions/RiskMetadata.yaml
@@ -74,7 +74,7 @@ properties:
     example:
       "User-Agent": "Mozilla/5.0"
       "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"
-  ipToBillingAddressDistance:
+  distance:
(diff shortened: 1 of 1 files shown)


spec/definitions/RiskMetadata.yaml
```diff
@@ -74,3 +74,7 @@ properties:
     example:
       "User-Agent": "Mozilla/5.0"
       "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"
+  ipToBillingAddressDistance:
+    description: Distance between IP Address and Billing Address geolocations (kilometers)
+    type: integer
+    readOnly: true
```
XaosSintez: I think the whole object is `readOnly` so you can set it there and don't care about every field.

spec/definitions/RiskMetadata.yaml
```diff
@@ -74,3 +74,7 @@ properties:
     example:
       "User-Agent": "Mozilla/5.0"
       "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"
+  ipToBillingAddressDistance:
```
XaosSintez: `ipToAddressDistance`?
`distance` and description which distance (as you have it)?

spec/definitions/RiskMetadata.yaml
viceromag: No, not RiskMetadata object - we can pass things like `httpHeaders` and `fingerprint`, it's the properties of ipAddress that are mostly read-only.

spec/definitions/RiskMetadata.yaml
viceromag: That was the original plan, but I thought it might be too vague... perhaps I'll bring that back on.

spec/definitions/RiskMetadata.yaml
```diff
@@ -74,3 +74,7 @@ properties:
     example:
       "User-Agent": "Mozilla/5.0"
       "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"
+  distance:
+    description: Distance between IP Address and Billing Address geolocations (kilometers)
```
slavcodev: `geolocations` -> `geolocation` ?

PR #369, opened 2017-10-21
Remove billingContact and deliveryContact links from Invoice

pmeth:
Remove `billingContact` and `deliveryContact` links from Invoice endpoints, as these are legacy fields. `billingAddress` and `deliveryAddress` are now fields within the invoice and not external.

### DIFF
--- a/spec/definitions/Invoice.yaml
+++ b/spec/definitions/Invoice.yaml
@@ -43,11 +43,11 @@ properties:
   billingAddress:
     description: Invoice's billing address
     allOf:
-      - $ref: "#/definitions/Contact"
+      - $ref: "#/definitions/ContactObject"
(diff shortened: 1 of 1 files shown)


spec/definitions/Invoice.yaml
```diff
@@ -43,11 +43,11 @@ properties:
   billingAddress:
     description: Invoice's billing address
     allOf:
-      - $ref: "#/definitions/ContactObject"
+      - $ref: "#/definitions/Contact"
```
XaosSintez: This is incorrect. This reference is for `Contact` entity which also has `id` and `customerId` (and some other fields) while `ContactObject` was created for handling contact related data inside resources

PR #374, opened 2017-11-02
Adjust fields according to API

2e3s:


2e3s:
I have reverted the arguable changes and added fixes for more obvious problem.

PR #392, opened 2017-12-07
Add AllowedAmountRanges to GatewayAccount

colinbird:


colinbird:
I think you are right about `amountRestrictions`

As for additional restrictions, we actually already have many..  currency, payment-method, payment-card-schemes, websites.  Should these all remain independent or be grouped somehow.  Perhaps because these will be optional restrictions

PR #394, opened 2017-12-13
Add 2 new endpoints to activate/deactivate gateway-accounts

Zaphell:


spec/paths/gateway-accounts@{id}@activate.yaml
```diff
@@ -0,0 +1,26 @@
+parameters:
+  - $ref: "#/parameters/resourceId"
+post:
+  tags:
+    - Gateway Accounts
+  summary: Activate a Gateway Account
+  description: |
+    Activate a Gateway Account with specified identifier string
+  responses:
+    201:
```
dara123: should it be `200` ?

spec/paths/gateway-accounts@{id}@activate.yaml
```diff
@@ -0,0 +1,26 @@
...
+          $ref: "#/headers/Rate-Limit-Limit"
+        Rate-Limit-Remaining:
+          $ref: "#/headers/Rate-Limit-Remaining"
+        Rate-Limit-Reset:
+          $ref: "#/headers/Rate-Limit-Reset"
+      schema:
+        $ref: "#/definitions/GatewayAccount"
+    401:
+      $ref: "#/responses/AccessForbidden"
+    422:
```
dara123: Do we have any validation for this?

spec/paths/gateway-accounts@{id}@activate.yaml
slavcodev: @dara123 no. it should be 201 and redirect to `/gateway-accounts@{id}`

Zaphell: Since this is POST, we want `201`

spec/paths/gateway-accounts@{id}@activate.yaml
Zaphell: We do not, we can however get a 404 :+1:

PR #399, opened 2017-12-19
Rename Checkout Page "uriPath" to "urlPathSegment"

creocoder:


adamaltman:
conflict

creocoder:
Checkout page spec was moved to separate repo, so we can safetly close that PR. It was just forgoten.

PR #400, opened 2017-12-20
Add "approvalWindowTtl" attribute to gateway account

Zaphell:
New functionality on Gateway Accounts to allow merchants to specify when can a transaction remain suspended until it is canceled automatically by Rebilly.

### DIFF
--- a/spec/definitions/GatewayAccount.yaml
+++ b/spec/definitions/GatewayAccount.yaml
@@ -62,6 +62,11 @@ properties:
     type: number
     format: double
     minimum: 0
+  suspendedTransactionLifespan:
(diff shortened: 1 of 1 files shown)


spec/definitions/GatewayAccount.yaml
```diff
@@ -62,6 +62,12 @@ properties:
     type: number
     format: double
     minimum: 0
+  approvalWindowTtl:
```
adamaltman: What about `suspendedTtl`?

Zaphell:
Replaced by: https://github.com/Rebilly/RebillyUserAPI/pull/3

PR #402, opened 2017-12-25
Email Notifications (proposal #1)

XaosSintez:
This PR is a proposal `#1` for Email Notifications project.

It is not really corresponds to the concept, but we discussed this approach too. And to see all the options, i'm preparing two PRs with proposals.

Definitions names and endpoints probably are not final.

I see these pros and cons for this approach:

Pros:

- It is flexible - it is possible to create many different notifications for different email addresses for same event. For example, customer can get simple notification, merchant can get another and CEO some other motification for same event, which can help them to process same order in different ways.
- It is more or less similar to global webhooks

Cons:

- It is not corresponding to the concept.

## TODO
Add trigger test notification

### DIFF
--- a/spec/definitions/EmailTemplate.yaml
+++ b/spec/definitions/EmailTemplate.yaml
@@ -1,0 +1,24 @@
+type: object
+required:
+  - subject
+  - body
+properties:
+  id:
+    description: The email template identifier string
+    readOnly: true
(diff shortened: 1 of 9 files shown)


XaosSintez:
Looks like this proposal does not fit requirements.

PR #404, opened 2018-01-03
add ListRestrictions

colinbird:


spec/definitions/ListRestrictions.yaml
```diff
@@ -0,0 +1,22 @@
...
+      type: string
+      enum:
+        - country
+        - bin
+  action:
+      description: How to treat the List items
+      type: string
+      enum:
+        - allow
+        - disallow
```
colinbird: `in` / `not-in` ?

Zaphell: With `in` / `not-in`, `condition` or `criteria` makes more sense than `action` as an attribute here... Having `allow` or `disallow` fits well with the `action` attribute though.
The endpoint `ListRestrictions` does however provide a clue as to what the action is, so maybe `criteria`/`condition`/`constraint` works well for that attribute.

What about `type` => `entity`, `action` => `type` (as in Restriction type is "in" or "not-in").

adamaltman:
Apologies, I think this should move to the RebillyUserApi spec.

PR #413, opened 2018-02-24
Issues an interim invoice

adamaltman:


spec/definitions/SubscriptionInvoice.yaml
```diff
@@ -0,0 +1,6 @@
+type: object
+properties:
+  paymentId:
```
colinbird: How would we know the amount for this payment, from the preview?

adamaltman: This is an id of a completed payment, maybe transactionId would be actually better.

adamaltman: (So the amount would come from that...)

PR #430, opened 2018-05-02
Fix payment card tokens description

adamaltman:


spec/definitions/PaymentToken.yaml
```diff
@@ -9,7 +9,8 @@ properties:
     allOf:
       - $ref: "#/definitions/ResourceId"
   method:
-    $ref: "#/definitions/Method"
+    - payment-card
+    - bank-account
```
Zaphell:
```
type: string
description: The payment method
enum:
  - payment-card
  - bank-account
```
Didn't we name `bank-account` as `ach`?

PR #435, opened 2018-05-07
Update authentication docs with support for publishable keys

xaben:
See preview here:
http://rebilly.github.io/RebillyAPI/preview/publishable_keys/


spec/swagger.yaml
```diff
@@ -34,24 +34,13 @@ info:
...
-
-    Rebilly also has a client-side authentication scheme that uses an
-    apiUser and HMAC-SHA1 signature (only for the Tokens resource), so
-    that you may safely create tokens from the client-side without compromising
-    your secret keys.
+    Rebilly offers four forms of authentication:  secret key, publishable key, JSON Web Tokens, and public signature key.
+    - [Secret API key](#section/Authentication/SecretApiKey): used for requests made from the server side. Never share these keys. Keep them guarded and secure
+    - [Publishable API key](#section/Authentication/PublishableApiKey): used for requests from the client side. For now can only be used on the [Tokens resource](#tag/Payment-Tokens%2Fpaths%2F~1tokens%2Fpost)
+    - [JWT](#section/Authentication/JWT): short lifetime tokens that can be assigned a specific expiration time
+    - [Public signature key](#section/Authentication/PublicSignatureKey): (deprecated) used for requests from the client side. For now can only be used on the [Tokens resource](#tag/Payment-Tokens%2Fpaths%2F~1tokens%2Fpost)
```
adamaltman: I think this should be removed.  If it is deprecated and we don't want to support it in the future, we shouldn't write about it.  As long as we publish it, we risk new integrations utilizing this method (even if it is marked `deprecated`).

PR #443, opened 2018-05-22
Adjust API specs with File Upload details and Publishable Key authentication

xaben:


spec/paths/files.yaml
```diff
@@ -49,19 +49,32 @@ post:
...
+    The following file types only are allowed:
+     - jpg
+     - png
+     - gif
+     - pdf
+     - mp3
+
+
+    If using a Publishable Api Key, only private files can be created. The files can later on be modified or used using
+     a private API key.
```
adamaltman: a `secret` API key.

spec/definitions/FileCreateRequest.yaml
```diff
@@ -0,0 +1,31 @@
+type: object
+required:
+  - file
```
adamaltman: I'm not sure this is accurate, but I don't think the version of OpenAPI we use allows for some conditionally required fields.

xaben: Well I added a message on the actual field description for that.

PR #447, opened 2018-05-30
Add orderId to Transaction

viceromag:


spec/definitions/Transactions/Transaction.yaml
```diff
@@ -283,6 +283,10 @@ properties:
     description: True if the transaction has been updated due to a disprepancy with its source of truth
     type: boolean
     readOnly: true
+  orderId:
+    description: The transactions's order ID
+    type: string
+    readOnly: true
```
slavcodev: It is going to be `writable`

PR #450, opened 2018-06-05
Add cancellation resource

2e3s:


spec/definitions/Cancellation.yaml
```diff
@@ -0,0 +1,120 @@
...
+      If empty then the invoice will be created.
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  policy:
+    description: Cancellation policy
+    type: string
+    enum:
+      - at-next-renewal
+      - at-specified-time
+  by:
```
slavcodev: please consider to use `cancelledBy`

2e3s: I think it's an obvious prefix being in cancellation resource

slavcodev: I do not think so.

XaosSintez: @slavcodev not `cancelledBy`, but `canceledBy` (with one `l`). And yes, `by` looks not very obvious

spec/definitions/Cancellation.yaml
```diff
@@ -0,0 +1,109 @@
...
+    description: Cancellation identifier
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  subscriptionId:
+    description: Identifier of the canceled subscription
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  invoiceId:
+    description: Identifier of the invoice to which the prorated amount is added.
+    readOnly: true
```
slavcodev: Is it read-only by requirements or design? I wonder may merchant want to set it explicitly?

spec/definitions/Cancellation.yaml
```diff
@@ -0,0 +1,109 @@
...
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  canceledBy:
+    description: Who did the cancellation
+    type: string
+    default: customer
+    enum:
+      - merchant
+      - customer
+  category:
```
slavcodev: Just wonder would be `reason` better name for this field? Or may be other. 
I feel that `category` might be confused, is it category of subscription, of the what?

slavcodev: I see it's clarifying in the description, but may be we can improve the field name too?

spec/paths/cancellations.yaml
```diff
@@ -0,0 +1,58 @@
...
+          $ref: "#/headers/Pagination-Limit"
+        Pagination-Offset:
+          $ref: "#/headers/Pagination-Offset"
+      schema:
+        type: array
+        items:
+          $ref: "#/definitions/Cancellation"
+    401:
+      $ref: "#/responses/AccessForbidden"
+post:
```
slavcodev: I see you added only `POST`, but I believe we should also support `PUT`. Because we have added the `preview` flag and it was a hack. Now with the resource for the cancellation, we can store resource and be able to modify its state, by changing the value of the `preview`. (Btw, we can consider to change this field name now)

spec/definitions/Cancellation.yaml
```diff
@@ -0,0 +1,109 @@
...
+    type: boolean
+    description: |
+      Defines if the resource isn't actually created but the response can be inspected so that the charge can be seen.
+    default: true
+  lineItems:
+    description: Items to be added to the next invoice
+    readOnly: true
+    type: array
+    items:
+      type: object
```
slavcodev: Is it different from `InvoiceItem`? or we can reuse the existing definition?

spec/definitions/Cancellation.yaml
adamaltman: Yes, good idea.

spec/definitions/Cancellation.yaml
```diff
@@ -0,0 +1,109 @@
...
+      Defines if the customer gets a pro-rata credit for the time remaining between "effectiveTime" and
+      subscription's next renewal time.
+    type: boolean
+    default: false
+  createdTime:
+    description: The cancellation time (when the resource is posted).
+    type: string
+    format: date-time
+    readOnly: true
+  effectiveTime:
```
adamaltman: This should be `churnTime`.

spec/definitions/Cancellation.yaml
```diff
@@ -0,0 +1,109 @@
...
+    description: The time when subscription will no longer be collected.
```
adamaltman: The time when the subscription will be deactivated.

spec/definitions/Cancellation.yaml
```diff
@@ -0,0 +1,109 @@
...
+properties:
+  id:
+    description: Cancellation identifier
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  subscriptionId:
+    description: Identifier of the canceled subscription
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  invoiceId:
```
adamaltman: I find this confusing.  There is the invoice which the proration is applied to.  

For example:
```
Invoice #1 

Plan Gold for June 1 to June 30th  $30
```

Canceled June 12th with proration and that means there should be a credit of $18.  The $18 is a credit against the $30 on `Invoice #1`.  

The credit may appear on `Invoice #2`.  

Which is this referring to?  Where are they?  A credit may be applied to multiple invoices.  How is that reflected here?

spec/definitions/Cancellation.yaml
```diff
@@ -0,0 +1,109 @@
...
+  effectiveTime:
+    description: The time when subscription will no longer be collected.
+    type: string
+    format: date-time
+  preview:
+    type: boolean
+    description: |
+      Defines if the resource isn't actually created but the response can be inspected so that the charge can be seen.
+    default: true
+  lineItems:
```
adamaltman: I don't understand the purpose of this except for the preview?  Is that correct?

2e3s: It is for a preview, yes

spec/definitions/Cancellation.yaml
2e3s: We can't reuse the existing definition because invoice item is a different resource/entity. However, this can be reused along with Subscription's line items (which are there for plan change preview).

spec/definitions/Cancellation.yaml
2e3s: This field is the identifier of the invoice to which the prorated amount is added which can be `Invoice #2` in your example. Do we need to know the last invoice with the full amount (`Invoice #1`)? To me the prorated cancellation amount is not directly related to `Invoice #1`, a customer pays for a new item of the partial subscription's period.
Or is such a basic name confusing?

spec/definitions/Cancellation.yaml
adamaltman: It doesn't make sense to me.

spec/definitions/Cancellation.yaml
2e3s: Actually I'm wrong.

spec/definitions/Cancellation.yaml
2e3s: I was wrong about the meaning of this invoice, it's the one on which the proration is calculated.
That what you're talking about could probably exist here too, but I'm not sure such feature is needed.

spec/paths/cancellations.yaml
adamaltman: You mean, like `status`?

2e3s: Status of cancellation "preview", which can be changed to "applied"?

slavcodev: Yes, I propose to have 2 states of the cancellation: pending/preview and completed

spec/definitions/Cancellation.yaml
```diff
@@ -0,0 +1,117 @@
...
+    description: Cancellation identifier
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  subscriptionId:
+    description: Identifier of the canceled subscription
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  canceledInvoiceId:
+    description: |
+      Identifier of the invoice on which the cancellation is applied, its total defines the proration amount.
```
adamaltman: What do you think: `The identifier of the invoice where the cancellation fees or credits are applied. `

spec/definitions/Cancellation.yaml
```diff
@@ -0,0 +1,117 @@
...
+      - contract-expired
+      - too-expensive
+      - other
+  description:
+    description: Cancel reason description in free form
+    type: string
+    maxLength: 255
+  prorated:
+    description: |
+      Defines if the customer gets a pro-rata credit for the time remaining between "effectiveTime" and
```
adamaltman: `churnTime`

spec/definitions/Cancellation.yaml
```diff
@@ -0,0 +1,117 @@
...
+      subscription's next renewal time.
+    type: boolean
+    default: false
+  status:
+    description: |
+      "draft" defines that the cancellation isn't applied on an invoice and subscription but
+      can be inspected to see the charge.
+      "confirmed" will set a subscription to be canceled eventually.
+      "completed" is a readonly status which is set when the subcription is canceled and cannot be changed.
```
adamaltman: `read-only`

spec/definitions/Cancellation.yaml
```diff
@@ -0,0 +1,117 @@
...
+    description: |
+      Defines if the customer gets a pro-rata credit for the time remaining between "effectiveTime" and
+      subscription's next renewal time.
+    type: boolean
+    default: false
+  status:
+    description: |
+      "draft" defines that the cancellation isn't applied on an invoice and subscription but
+      can be inspected to see the charge.
+      "confirmed" will set a subscription to be canceled eventually.
```
adamaltman: instead of `eventually`, when the `churnTime` is reached.

spec/definitions/Cancellation.yaml
slavcodev: Did we already support `cancellation fees`? I think no, and this part seems might be confused.

spec/definitions/Cancellation.yaml
```diff
@@ -0,0 +1,117 @@
...
+    type: string
+    default: customer
+    enum:
+      - merchant
+      - customer
+  reason:
+    description: Cancellation category
+    type: string
+    default: other
+    enum:
```
adamaltman: If I recall, there is a category called `billing-failure` too.

spec/definitions/Cancellation.yaml
```diff
@@ -0,0 +1,117 @@
...
+      - $ref: "#/definitions/ResourceId"
+  canceledBy:
+    description: Who did the cancellation
+    type: string
+    default: customer
+    enum:
+      - merchant
+      - customer
+  reason:
+    description: Cancellation category
```
slavcodev: reason

spec/definitions/Cancellation.yaml
```diff
@@ -0,0 +1,117 @@
...
+        periodStartTime:
+          description: Date-time when the period begins for this item
+          type: string
+          format: date-time
+        periodEndTime:
+          description: Date-time when the period ends for this item
+          type: string
+          format: date-time
+        createdTime:
+          description: Date-time when the item was added to the subscription
```
adamaltman: This item isn't added to the subscription.

spec/definitions/Cancellation.yaml
```diff
@@ -0,0 +1,117 @@
...
+    type: string
+    format: date-time
+    readOnly: true
+  churnTime:
+    description: The time when the subscription will be deactivated.
+    type: string
+    format: date-time
+  lineItems:
+    description: Items to be added to the new invoice
+    readOnly: true
```
adamaltman: how can someone add a free-form cancellation fee, for example, if this is `readOnly`?  Or override some other fee?

spec/definitions/Cancellation.yaml
adamaltman: I think this needs to be more clear:

"completed" is a read-only status which is set by the system when the `churnTime` is reached.  The cancellation may not be changed or deleted when the status is "completed".

spec/definitions/Cancellation.yaml
2e3s: Yes, if I were a client and read this description, I would go to look how I can setup cancellation fees.

spec/definitions/Cancellation.yaml
2e3s: I don't think that `createdTime` is needed here at all. It's always cancellation time anyway.

spec/definitions/Cancellation.yaml
2e3s: However, if `lineItems` is not read-only then a fee can be added...

spec/definitions/Cancellation.yaml
2e3s: Not sure it would look nice considering that some items will be autogenerated. But ok.

slavcodev: May be leave these lines read-only? and make writeable `appliedInvoiceId ` so merchant can use pre-composed invoice with all fees?

spec/definitions/Cancellation.yaml
```diff
@@ -0,0 +1,118 @@
...
+      "completed" is a read-only status which is set by the system when the churnTime is reached.
+      The cancellation may not be changed or deleted when the status is "completed".
+    type: string
+    default: confirmed
+    enum:
+      - draft
+      - confirmed
+      - completed
+  createdTime:
+    description: The cancellation time (when the resource is posted).
```
adamaltman: As you pointed out, it may be a draft cancellation, and until it is confirmed, it's not really canceled.  Therefore, we should also have a canceledTime.

adamaltman:
Missing `_links`?

PR #452, opened 2018-06-08
Update payment card spec

dara123:


spec/paths/payment-cards@{id}.yaml
```diff
@@ -39,9 +39,25 @@ patch:
           cvv:
             description: Card's cvv (card verification value).
             type: string
+            required: false
+          expMonth:
+            description: Card's expired month
```
adamaltman: Card's expiry month

spec/paths/payment-cards@{id}.yaml
```diff
@@ -39,9 +39,25 @@ patch:
...
+            type: string
+            required: false
+          expYear:
+            description: Card's expired year
```
adamaltman: Card's expiry year

spec/paths/payment-cards@{id}.yaml
```diff
@@ -39,9 +39,25 @@ patch:
           cvv:
             description: Card's cvv (card verification value).
             type: string
+            required: false
```
adamaltman: required is false by default, so why include it here?

slavcodev: actually is not supported by swagger, `required` is attribute of the `object` that contains list of required properties, is not a flag for every field

adamaltman: Yes, I think this is why the tests are failing.

spec/paths/payment-cards.yaml
```diff
@@ -43,6 +43,17 @@ post:
       schema:
         $ref: "#/definitions/PaymentCard"
   responses:
+    200:
+      description: PaymentCard was updated
```
adamaltman: isn't this created?

adamaltman: oh nevermind, I understand.

PR #460, opened 2018-06-15
Add KYC documents spec

xaben:


spec/definitions/KycDocument.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - $ref: "#/definitions/ResourceId"
+  customerId:
+    description: Linked Customer object id
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  documentType:
+    description: Document type submitted for validation, only identity_proof type is analyzed in an automated manner
+    type: string
+    enum:
+      - identity_proof
```
slavcodev: We use `kebab case`

spec/definitions/KycDocumentMatches.yaml
```diff
@@ -0,0 +1,51 @@
+type: object
+readOnly: true
+properties:
+  version:
```
slavcodev: Is it useful for public?

spec/definitions/KycDocumentMatches.yaml
```diff
@@ -0,0 +1,51 @@
...
+          - $ref: "#/definitions/ServerTimestamp"
+      expiryDate:
+        description: The expiry date found on the document, null if not found
+        allOf:
+          - $ref: "#/definitions/ServerTimestamp"
+      issueDate:
+        description: The issued date found on the document, null if not found
+        allOf:
+          - $ref: "#/definitions/ServerTimestamp"
+      age:
```
slavcodev: I believe this is unnecessary value. The client should calculate this itself.

spec/definitions/Links/KycDocumentLink.yaml
```diff
@@ -0,0 +1,11 @@
+type: object
```
slavcodev: Where is this used?

spec/paths/kyc-documents.yaml
```diff
@@ -0,0 +1,79 @@
...
+        - items:
+            x-sortableEnum:
+              - id
+              - createdTime
+              - updatedTime
+    - x-rebillyMerge:
+      - $ref: "#/parameters/mediaType"
+      - enum:
+        - application/json
+        - text/csv
```
slavcodev: Are we going to support this?

spec/paths/kyc-documents@{id}@acceptance.yaml
```diff
@@ -0,0 +1,24 @@
+parameters:
+  - $ref: "#/parameters/resourceId"
+post:
+  tags:
+    - KYC Documents
+  summary: Accept a KYC document
+  description: |
+    Manually override automated status and accept the document
+  responses:
+    200:
```
slavcodev: should be `201`

spec/paths/kyc-documents@{id}@refusal.yaml
```diff
@@ -0,0 +1,35 @@
...
+    Manually override automated status and refuse the document
+  parameters:
+    - name: body
+      in: body
+      description: KYC document resource
+      required: true
+      schema:
+        $ref: "#/definitions/KycDocumentRefusal"
+  responses:
+    200:
```
slavcodev: should be `201`

spec/definitions/KycDocumentMatches.yaml
slavcodev: One of the reason the `cache`. Such fields calculated on every request, just kill the one of principe of the REST.

spec/paths/kyc-documents@{id}@acceptance.yaml
slavcodev: This created another resource `document acceptance`. If you want to update the existing `document`, we should use `PUT /kyc-documents/{id}`

spec/paths/kyc-documents@{id}@refusal.yaml
slavcodev: Same as with `acceptance`.

spec/definitions/KycDocumentMatches.yaml
adamaltman: The age only changes once every year though.

spec/paths/kyc-documents@{id}@acceptance.yaml
adamaltman: Hmm, I think I agree and disagree with @slavcodev ... This probably should create another resource.  Although, I'm not sure if something can be accepted and refused (I think we should rename that to `rejected` btw, I'll post a separate comment about that).

Can something be accepted multiple times?  It seems the kyc document has a status, which is the most recent decision.  However, their may be multiple decisions (a decision history).

spec/definitions/KycDocument.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - address-proof
+  status:
+    description: Status of the validation
+    type: string
+    readOnly: true
+    enum:
+      - pending
+      - in-progress
+      - approved
+      - refused
```
adamaltman: I think `refused` is the wrong word to use.  `Rejected` is the correct antonym to use.

spec/definitions/KycDocument.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - identity-proof
+      - address-proof
+  status:
+    description: Status of the validation
+    type: string
+    readOnly: true
+    enum:
+      - pending
+      - in-progress
+      - approved
```
adamaltman: Below you have an endpoint named `acceptance`, and this would seem to make more sense if the status here was `accepted`.  On the other hand, it would make more sense if the endpoint was named `approvals` if the status is `approved`.

spec/definitions/KycDocument.yaml
```diff
@@ -0,0 +1,59 @@
...
+  fileId:
+    description: Linked File object id
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  customerId:
+    description: Linked Customer object id
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  documentType:
+    description: Document type submitted for validation, only identity_proof type is analyzed in an automated manner
```
adamaltman: Incorrect `identity_proof` should use kebab-case.

spec/definitions/KycDocumentMatches.yaml
slavcodev: Yes, but anyways, let imagine the client requests the resource one before DOB.
And there are a cache between backend and client, I thin it would be annoyed to cache invalidation for this resource, so that on next day cache will be invalidated and client will get fresh data.

Of course this field can be in resource, and we can support it by any means, but I do not see a reason for it. I think we underestimate the capabilities of clients. :)

spec/paths/kyc-documents@{id}@acceptance.yaml
slavcodev: By `new resource` I meant the usage of new `URL`. If we doing `POST` and backend respond with `200`, that means for client that payload accepted and was modified resource specified by URL, that means the client can request `GET` this URL and get the new state of the resource, which will not work in this case. `201` allows us to redirect the client to the resource that client can work further,  using `Location`.

adamaltman:
@slavcodev brings up an interesting thoughts.  Are the approvals and rejections resources?  Are they different resources?  Or is there an analysis resource?  

@cesarlevel brought an interesting issue on Friday.  Should the approvals and rejections have some user id/name?  They need to be associated with a user to achieve the user interface goals.

spec/definitions/KycDocumentMatches.yaml
slavcodev: > I think you overestimate it. Because first, we have no clients that actually use caching. Side note: We also have no clients that use our hyperlinks either.

You're right, we do not have such clients. I just when think about API I think about all this things. I do not think we want to change our API after one month again.

> Second, I thought about this more. What is important is the age at the time of the analysis

That sounds interesting. I think I agree that the age makes sense on moment when doc is analysed. However that is other requirements, and perhaps we should somehow change the name of the fields to reflect this details.

xaben: Ok, perhaps we should rename the field in this case. I am thinking perhaps we can calculate it as the diff between the date of birth and createdTime for the document, instead of storing it?

spec/definitions/KycDocumentMatches.yaml
xaben: Yeah I guess we can hide that.

spec/paths/kyc-documents.yaml
xaben: This is something we have on all resources as far as I know. I wasn't sure if the serializer somehow transforms this based on the Accept header. I will check.

spec/paths/kyc-documents@{id}@acceptance.yaml
xaben: Yes it is 201 on the backend, so I will adjust it. As for now we don't have a decision history, perhaps it would be a good thing though, should I add a ticket?

spec/definitions/KycDocument.yaml
xaben: Ok will change that

PR #458, opened 2018-06-14
Api spec for Visa validation

dara123:

PR #463, opened 2018-06-18
Revert "Api spec for Visa validation"

dara123:
This make Rebilly test fail becuase couldn't find this new endpoint yet. 
Will merge this PR new API spec when we implement this new endpoint

Reverts Rebilly/RebillyAPI#458


dara123:
Revert this for now because it makes Rebilly test failed

dara123:
dev please review

PR #466, opened 2018-06-26
Add reactivation resource

2e3s:
Related to #450


spec/definitions/Reactivation.yaml
```diff
@@ -0,0 +1,34 @@
...
+    allOf:
+      - $ref: "#/definitions/ResourceId"
+  reactivatedBy:
+    description: Who did the reactivation
+    type: string
+    default: customer
+    enum:
+      - merchant
+      - customer
+  description:
```
adamaltman: reason may be more clear.

2e3s:
@adamaltman no, this was my intiative to add this field. And I decided to remove it because it is probably not in demand.

spec/definitions/Reactivation.yaml
2e3s: We have description in

2e3s: We have "description" for a free form text in the cancellation, I made it so to be consistent

2e3s:
About a default value.. isn't it that the less API requires the easier it is to use?

adamaltman:
I don't think so.  Not when it has mystery defaults.

adamaltman:
Missing _links with rel self, and get one endpoint.

2e3s:
Actually, this specification is already implemented.

PR #467, opened 2018-06-27
Add invoice send action

vova07:


vova07:
please review

spec/paths/invoices@{id}@send.yaml
```diff
@@ -0,0 +1,26 @@
+parameters:
+  - $ref: "#/parameters/resourceId"
+post:
+  tags:
+    - Invoices
+  summary: Void an invoice
```
2e3s: void?

spec/paths/invoices@{id}@send.yaml
```diff
@@ -0,0 +1,26 @@
...
+      description: Invoice was sent successfully
+      headers:
+        Rate-Limit-Limit:
+          $ref: "#/headers/Rate-Limit-Limit"
+        Rate-Limit-Remaining:
+          $ref: "#/headers/Rate-Limit-Remaining"
+        Rate-Limit-Reset:
+          $ref: "#/headers/Rate-Limit-Reset"
+      schema:
+        $ref: "#/definitions/Invoice"
```
XaosSintez: `204` means that there's no response

spec/paths/invoices@{id}@send.yaml
```diff
@@ -0,0 +1,26 @@
+parameters:
+  - $ref: "#/parameters/resourceId"
+post:
+  tags:
+    - Invoices
+  summary: Void an invoice
+  description: |
+    Void an invoice with specified identifier string
```
XaosSintez: I think you forgot to change this

spec/paths/invoices@{id}@send.yaml
vova07: Wrong copy-paste!

spec/paths/invoices@{id}@send.yaml
```diff
@@ -0,0 +1,26 @@
...
+          $ref: "#/headers/Rate-Limit-Remaining"
+        Rate-Limit-Reset:
+          $ref: "#/headers/Rate-Limit-Reset"
+      schema:
+        $ref: "#/definitions/Invoice"
+    401:
+      $ref: "#/responses/AccessForbidden"
+    404:
+      $ref: "#/responses/NotFound"
+    422:
```
XaosSintez: What kind of invalid data can be here?

spec/paths/invoices@{id}@send.yaml
vova07: Yes!
Should we return anything back?

spec/paths/invoices@{id}@send.yaml
vova07: Invalid `EmailNotification` attributes.

vova07: To be clear on it:
The idea is to accept an `email template` which will be processed future.
So in case of invalid data, we should throw 422 error.

XaosSintez: But there's no body specified

spec/paths/invoices@{id}@send.yaml
```diff
@@ -0,0 +1,26 @@
+parameters:
+  - $ref: "#/parameters/resourceId"
+post:
+  tags:
+    - Invoices
+  summary: Send an invoice
+  description: |
+    Send an invoice via email with specified identifier string
```
2e3s: via which email?

vova07: Not sure I got the point.
I mean it will be send as an `email`.
Isn't it right to use here `via`?

spec/paths/invoices@{id}@send.yaml
vova07: Done!

spec/paths/invoices@{id}@send.yaml
2e3s: Which email will be used? I don't understand it from the description

vova07: If you mean from which email address it will be sent, then the one which will be provided in the body params.

spec/paths/invoices@{id}@send.yaml
```diff
@@ -0,0 +1,33 @@
+parameters:
```
slavcodev: I don't understand this action.
Client should use `EmailNotification` to send invoice? When immediately on request?
EmailNotification has `eventType`? How it will work?
What about `status`?

vova07: I'll like to switch to implementation PR, where I'll can provide more details.

spec/paths/invoices@{id}@send.yaml
adamaltman: From which sender?

PR #468, opened 2018-07-01
Fix cancellations path

2e3s:


adamaltman:
There is another issue I had in my local, at least.  I needed to change a plan to be a 30 day plan instead of a 1 month plan, because month size varies, your pro rata test was failing in this month.... I suppose it already is fixed somwhere if tests passed here.

PR #474, opened 2018-07-21
Adjust plans for pricing formulas

adamaltman:


spec/definitions/Plan.yaml
```diff
@@ -45,10 +51,50 @@ properties:
...
+      properties:
+        price:
+          description: If the price is 0, it's free.
+          type: number
+          format: double
+          example: 99.95
+        maxQuantity:
+          description: |
+            If the value is `null`, it's assumed to be infinite.  It starts at the end of the previous bracket.
+            If there is no previous bracket, it starts at 1.
```
colinbird: This is a bit confusing but I think I understand it, maybe an example would help

spec/definitions/Plans/Plan.yaml
```diff
@@ -19,50 +20,55 @@ properties:
...
-  trialPeriodLength:
-    description: The length of time (used with the trialPeriodUnit)
-    type: integer
-  setupAmount:
-    description: The amount of a setup - 0 is a valid value (for free)
-    type: number
-    format: double
+  pricing:
+    $ref: "#/definitions/PlanPriceFormula"
+  recurringInterval:
```
2e3s: It is inconsistent. In trial it is called "period", here "interval".

spec/definitions/Plans/Plan.yaml
```diff
@@ -19,50 +20,55 @@ properties:
...
+  trial:
+    type: object
+    description: The trial. Set `null` if no trial.
+    required:
+      - price
+      - period
+    properties:
+      price:
+        description: The price of the trial. For a free trial, use `0`.
+        type: number
```
2e3s: Will the trial be priced in a complex way?

2e3s:
Would there be a field to denote a variable usage?

spec/definitions/Plans/Plan.yaml
slavcodev: Well, it's not supposed to be consistent.
Trial - is a period of time when product is provided for special price.
The main price is billed in intervals, isn't it?

spec/definitions/Plans/Plan.yaml
slavcodev: I do not know. Should it be? We've not discussed such requirement.

slavcodev:
> Would there be a field to denote a variable usage?

Yes. But the changes related to the `Usage AP` will be in different PR, just to not block the release with plan pricing.

spec/definitions/Plans/Plan.yaml
2e3s: To me "period" and "interval" are just synonyms

spec/definitions/Plans/Plan.yaml
2e3s: I don't know. I can imagine a trial not being flat rate but just reduced recurring price (which may be complex). Probably this should be discussed before going forward.

spec/definitions/Plans/Plan.yaml
slavcodev: are they?

2e3s: More strictly saying, a period is an interval of time.

2e3s: To me it's okay to name `recurring` by analog of trial

2e3s: And maybe have `pricing` with this object?

slavcodev: I do not think so. For me `recurring interval` and `pricing` are two different things, for example you can have `price` w/o specified `recurring interval`

spec/definitions/Plans/Plan.yaml
adamaltman: No.

slavcodev:
devs please review

PR #475, opened 2018-07-25
Make Subscription support multiple plans

2e3s:


spec/definitions/Subscription.yaml
```diff
@@ -13,10 +13,19 @@ properties:
...
       - $ref: "#/definitions/ResourceId"
-  planId:
-    description: Unique id for each plan
-    allOf:
-      - $ref: "#/definitions/ResourceId"
+  items:
+    type: array
+    minItems: 1
+    items:
+      type: object
```
XaosSintez: Are `planId` and `quantity` required inside this object?

adamaltman: `planId` definitely is, I think.

2e3s: Maybe quantity may be omitted (null)

PR #476, opened 2018-07-26
Rename hasVariableUsage to prepaid 

slavcodev:
Relates to #474


slavcodev:
devs please review

spec/definitions/Plans/Plan.yaml
```diff
@@ -22,6 +22,10 @@ properties:
   isActive:
     description: If the plan is not active, customers cannot subscribe to the plan.
     type: boolean
+  prepaid:
+    description: False if the plan will be billed in arrears and usage data will be supplied to calculate the quantity. Defaults to `true`.
```
2e3s: `and usage data will be supplied to calculate the quantity` is misleading, I guess. Why have it? If it's only about when this is billed. It may be mentioned as an example.
Also ```Defaults to `true`.``` is excessive

2e3s:
It looks more straightforward to me

2e3s:
But how do we know if the plan is usage based?

spec/definitions/Plans/Plan.yaml
```diff
@@ -85,9 +90,6 @@ properties:
             If there is no previous bracket, it starts at 1.
           type: integer
           example: 1
-  hasVariableUsage:
```
adamaltman: I think we still need to know if a plan has metered usage.

slavcodev: OK I will add it back in other PR, picking the name.

PR #478, opened 2018-07-27
Refactor Plan trial and setup properties

slavcodev:
Proposal for https://github.com/Rebilly/RebillyAPI/pull/474

### DIFF
--- a/spec/definitions/Plans/Plan.yaml
+++ b/spec/definitions/Plans/Plan.yaml
@@ -91,25 +91,29 @@ properties:
   hasVariableUsage:
     description: True if the plan will be billed in arrears and usage data will be supplied to calculate the quantity.
     type: boolean
-  trialPrice:
-    description: The price of the trial.  For a free trial, use `0`.
-    type: number
-    format: double
-  trialPeriodUnit:
-    description: The unit of time
-    type: string
-    enum:
-      - day
-      - week
-      - month
-      - year
-  trialPeriodLength:
-    description: The length of time (used with the trialPeriodUnit)
-    type: integer
-  setupPrice:
-    description: The price of a setup - 0 is a valid value (for free)
-    type: number
-    format: double
+  trial:
+    type: object
+    description: The trial. Set `null` if no trial.
+    allOf:
+      - type: object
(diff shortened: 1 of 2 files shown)


slavcodev:
devs please review

spec/definitions/Plans/Plan.yaml
```diff
@@ -91,25 +91,29 @@ properties:
...
+    allOf:
+      - type: object
+        required:
+          - price
+        properties:
+          price:
+            description: The price of the trial. For a free trial, use `0`.
+            type: number
+            format: double
+      - $ref: "#/definitions/PlanPeriod"
```
2e3s: Why did you move this out?

spec/definitions/Plans/PlanPeriod.yaml
```diff
@@ -0,0 +1,16 @@
+type: object
+required:
```
2e3s: looks strange

slavcodev: Yeah, very strange. In my editor it correct: `type`, `required`, etc.

spec/definitions/Plans/Plan.yaml
slavcodev: To make specification more explicit. In previous version you can specify trial price w/o to specify trial period. In new version (since we going to change considerably the definition) you can't do that.

2e3s: No, I support the change as I see it, but I meant that why to move the definition in a separate file when it's not reused.

slavcodev: Ahh, it's supposed to be reused for `recuring`.

spec/definitions/Plans/PlanPeriod.yaml
slavcodev: Seems something wrong only with how this diff is shown, the file looks ok
https://github.com/Rebilly/RebillyAPI/blob/8b94b87772ab80150b65df3871a2e9277e3dc9f3/spec/definitions/Plans/PlanPeriod.yaml

PR #482, opened 2018-08-01
Draft subscription

dara123:


dara123:
devs please review

spec/definitions/Subscription.yaml
```diff
@@ -140,7 +140,7 @@ properties:
     allOf:
       - $ref: "#/definitions/RiskMetadata"
   startTime:
-    description: Subscription start time
+    description: You can have subscription start on a future date. If this field is set in the future, subscription is created now but will not issue invoice until future date.
```
colinbird: "You can have the subscription start in the future. If the given startTime is in the future, the subscription is created now but no invoices will be issued until the given startTime."

spec/definitions/Subscription.yaml
```diff
@@ -152,7 +152,7 @@ properties:
     allOf:
       - $ref: "#/definitions/ServerTimestamp"
   renewalTime:
-    description: Subscription renewal time
+    description: You can choose a specific day to do rebill. For example 1st of the month and prorated maybe apply.
```
colinbird: Pro-rated applies if set change this?

colinbird: "You can choose a specific date to rebill. For example, setting renewalTime to the 1st of the next month.  If `renewalPolicy` is set to `retain`, then a pro rata debit will occur."

I am not sure about this, is it right?

spec/definitions/Subscription.yaml
```diff
@@ -167,6 +167,12 @@ properties:
     description: Subscription updated time
     allOf:
       - $ref: "#/definitions/ServerTimestamp"
+  prorated:
+    description: |
+      Whether or not to give a pro rata credit for the amount of time remaining between the `effectiveTime` and the end of the current period.
+      In addition, if the `renewalTime` is retained (by setting the `renewalPolicy` to `retain`), then a pro rata debit will occur as well,
+      for the amount between the `effectiveTime` and the `renewalTime` as a percentage of the normal period size.
+    type: boolean
```
colinbird: should we say "prorated" instead of "pro rata" ?

spec/definitions/Subscription.yaml
dara123: I think just need `prorated = true`

spec/definitions/Subscription.yaml
```diff
@@ -152,7 +152,7 @@ properties:
     allOf:
       - $ref: "#/definitions/ServerTimestamp"
   renewalTime:
-    description: Subscription renewal time
+    description: You can choose a specific date to rebill. For example, setting renewalTime to the 1st of the next month. If prorated is set to true, then a prorated credit may occur.
```
2e3s: Maybe make these strings multiline?

spec/definitions/Subscription.yaml
```diff
@@ -34,10 +34,10 @@ properties:
     allOf:
       - $ref: "#/definitions/ContactObject"
   status:
-    description: Subscription status is deprecated and the values will change to `active`, `canceled`.
+    description: Subscription status. Can only set to "Draft" when you create a new subscription and we won't issue invoice or collect payment until it is activated.
```
2e3s: what can be done after `draft`?

spec/definitions/Subscription.yaml
```diff
@@ -34,10 +34,10 @@ properties:
...
     type: string
-    readOnly: true
     enum:
+      - Draft
```
2e3s: If we change the values and it will be `draft`, `active`, `canceled`, then maybe it's better to put it as `draft`.

spec/definitions/Subscription.yaml
dara123: we can activate after draft

2e3s: How?

2e3s: It says 

> Can only set to "Draft" when you create a new subscription

dara123:
devs please review

spec/definitions/Subscription.yaml
```diff
@@ -167,6 +167,10 @@ properties:
     description: Subscription updated time
     allOf:
       - $ref: "#/definitions/ServerTimestamp"
+  prorated:
```
slavcodev: How this is related to subscription drafts?

spec/definitions/Subscription.yaml
```diff
@@ -167,6 +167,10 @@ properties:
...
+    description: |
+      Whether or not to give a prorated credit for the amount of time remaining between the `effectiveTime` and the end of the current period.
+    type: boolean
```
slavcodev: It seems it's not required, does it have defaults?

spec/definitions/Subscription.yaml
```diff
@@ -34,10 +34,10 @@ properties:
     allOf:
       - $ref: "#/definitions/ContactObject"
   status:
-    description: Subscription status is deprecated and the values will change to `active`, `canceled`.
+    description: Subscription status. Can only set to "draft" when you create a new subscription and we won't issue invoice or collect payment until it is activated.
```
slavcodev: How it should be activated?

spec/definitions/Subscription.yaml
colinbird: It is not, I had the same question, I think that's why this PR may be confusing

spec/definitions/Subscription.yaml
dara123: it's default to `false`

spec/definitions/Subscription.yaml
dara123: https://docs.google.com/document/d/1ON6FJjkiTE5yebn5Vqz6GD533lrw4TpFSMzZgHb61Jc/edit#heading=h.f6kdf2r9zq2g

spec/definitions/Subscription.yaml
dara123: let say they draft a subscription (recur every month) on 15 and they set renewalTime on 1st we need to know if they want to do prorated

slavcodev: Alright, I just meant that seems different issues: one is making `renewalTime` changeable with `prorated` property and another feature is drafting subscriptions. Perhaps, I'm not sure though.

spec/definitions/Subscription.yaml
slavcodev: It's missing in the specs.

dara123:
dup: https://github.com/Rebilly/RebillyAPI/pull/483

PR #483, opened 2018-08-06
Draft subscription

dara123:


dara123:
devs please review

spec/definitions/Subscription.yaml
```diff
@@ -152,7 +152,7 @@ properties:
     allOf:
       - $ref: "#/definitions/ServerTimestamp"
   renewalTime:
-    description: Subscription renewal time
+    description: You can choose a specific date to rebill. For example, setting renewalTime to the 1st of the next month. If prorated is set to true, then a prorated credit may occur.
```
colinbird: i guess this description doesn't apply since prorated has been removed

dara123: removed

spec/definitions/Subscription.yaml
```diff
@@ -34,10 +34,10 @@ properties:
     allOf:
       - $ref: "#/definitions/ContactObject"
   status:
-    description: Subscription status is deprecated and the values will change to `active`, `canceled`.
+    description: Subscription status. Can only set to "draft" when you create a new subscription and we won't issue invoice or collect payment until it is activated.
```
colinbird: I think it's better to use backticks in descriptions for things like `draft`, and below it can be used for `startTime`, `renewalTime` ?

spec/definitions/Subscription.yaml
```diff
@@ -34,10 +34,10 @@ properties:
     allOf:
       - $ref: "#/definitions/ContactObject"
   status:
-    description: Subscription status is deprecated and the values will change to `active`, `canceled`.
+    description: Subscription status. Can only set to `draft` when you create a new subscription and we won't issue invoice or collect payment until it is activated.
```
slavcodev: Can you please re-add the notice about deprecated values?

spec/definitions/Subscription.yaml
```diff
@@ -140,7 +140,7 @@ properties:
     allOf:
       - $ref: "#/definitions/RiskMetadata"
   startTime:
-    description: Subscription start time
+    description: You can have the subscription start in the future. If the given startTime is in the future, the subscription is created now but no invoices will be issued until the given startTime.
```
slavcodev: I think the attributes names we should always put in the backticks.

PR #484, opened 2018-08-06
Add prorated to subscription

dara123:


dara123:
devs please review

spec/definitions/Subscription.yaml
```diff
@@ -155,6 +155,11 @@ properties:
     description: Subscription renewal time
     type: string
     format: date-time
+  prorated:
+    description: |
+      Whether or not to give a prorated credit for the amount of time remaining between the `effectiveTime` and the end of the current period.
```
colinbird: what is `effectiveTime` ?

2e3s: It's a copypaste from cancellations maybe. It should be "time between now and next renewalTime", probably. If I understand the feature correctly.

PR #487, opened 2018-08-09
Update spec payment card patch status

dara123:


dara123:
devs please review

spec/paths/payment-cards@{id}.yaml
```diff
@@ -44,6 +44,12 @@ patch:
           expYear:
             description: Card's expiry year
             type: string
+          status:
+            description: Card's status
+            type: string
+            enum:
+              - active
+              - deactivated
```
colinbird: should we add `inactive` ?

wmdmgv: @colinbird the same thoughts ... but maybe deactivated instead of inactive ? 
@dara123 what is the right?

PR #496, opened 2018-08-27
Add original fields to LeadSource

viceromag:


spec/definitions/LeadSource.yaml
```diff
@@ -1,10 +1,59 @@
 type: object
 properties:
-  id:
-    description: The lead source identifier string
+  originalMedium:
```
2e3s: I suggest to have `original: {medium: "x", ...}`

2e3s: much easier to work with such API IMO

viceromag: @2e3s what about recent lead source? Have it as ```recent: {medium: "x", ...}```? In that case, we won't be supporting any BC

slavcodev: I would even suggest the second properties in owners, i.e.
```
leadSource:
  $ref: "#/definitions/LeadSource"
originalLeadSource:
  $ref: "#/definitions/LeadSource"
```

2e3s: @viceromag yes, it could be `recent: {...}`, but I guess you will have to deal with BC so unprefixed values will exist anyway for some time, so I didn't mention them (it could be done separately).

PR #503, opened 2018-09-11
Add subscription count to plans

2e3s:
The value already exists in API response. It would make unnecessary 1 request for frontend.

### DIFF
--- a/spec/definitions/Plans/Plan.yaml
+++ b/spec/definitions/Plans/Plan.yaml
@@ -78,7 +78,7 @@ properties:
   subscriptionCount:
     type: integer
     readOnly: true
-    description: Number of subscriptions which this plan has
+    description: Number of subscriptions associated with this plan
   createdTime:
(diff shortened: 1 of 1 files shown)


spec/definitions/Plans/Plan.yaml
```diff
@@ -75,6 +75,10 @@ properties:
   richDescription:
     description: The plan rich description - supports HTML (deprecated)
     type: string
+  subscriptionCount:
+    type: integer
+    readOnly: true
+    description: Number of subscriptions which this plan has
```
adamaltman: including canceled subscriptions?

adamaltman: > Number of subscriptions associated with this plan.

2e3s: Yes @adamaltman the number includes all subscriptions with no condition

PR #526, opened 2018-10-24
Update and rename SubscriptionReactivation

slavcodev:


spec/definitions/Subscription/SubscriptionReactivation.yaml
```diff
@@ -20,9 +20,16 @@ properties:
     description: Reactivation reason description in free form
     type: string
     maxLength: 255
+  effectiveTime:
+    description: |
+      The date from which the service period would start, unless the suscbription is cancelled but still active.
+      If case the susbcription is stilla ctive, the susbcription will continue current service period.
```
2e3s: Type `stilla ctive`

adamaltman:
```suggestion
      In case the susbcription is still active, the subscription will continue the current service period.
```

spec/definitions/Subscription/SubscriptionReactivation.yaml
```diff
@@ -20,9 +20,16 @@ properties:
     description: Reactivation reason description in free form
     type: string
     maxLength: 255
+  effectiveTime:
+    description: |
+      The date from which the service period would start, unless the suscbription is cancelled but still active.
```
adamaltman:
```suggestion
      The date from which the service period would start, unless the subscription is canceled but still active.
```

PR #535, opened 2018-11-15
Fix references and small typos

ashkarpetin:
Adjusted references so that corresponding fields will not be ignored and corrected small typos.


ashkarpetin:
devs please review

PR #537, opened 2018-11-19
Update Bank account api spec

dara123:


spec/definitions/BankAccount.yaml
```diff
@@ -13,17 +13,26 @@ properties:
...
+    description: Bank's Routing Number. Required if bank account is not created from Token for BBAN account number type. This field is write-only
     type: string
   accountNumber:
     description: Bank's Account Number. Required if bank account is not created from Token. This field is write-only
     type: string
   accountType:
-    description: Banks's Account type. Required if bank account is not created from Token
+    description: Banks's Account type. Required if bank account is not created from Token for BBAN account number type.
     type: string
+  accountNumberType:
```
slavcodev: Perhaps would be better to put before `accountNumber`

spec/definitions/BankAccount.yaml
```diff
@@ -13,17 +13,26 @@ properties:
...
+    description: Bank's Account Number type. A valid value is basic bank account number (BBAN) or international bank account number (IBAN)
```
slavcodev: Is it read-only or writable?

slavcodev: Any defaults if writable?

dara123: writable

dara123: default to BBAN

slavcodev: please define the defaults then

PR #539, opened 2018-11-23
Add Subscription property orderType

slavcodev:
# Details
Adds the discriminator `orderType`. However it required to split the Subscription definition in multiple pieces in order to better sort the properties.

### DIFF
--- a/spec/definitions/Subscription.yaml
+++ b/spec/definitions/Subscription.yaml
@@ -1,16 +1,39 @@
 type: object
+discriminator: "orderType"
 required:
   - customerId
-  - items
   - websiteId
+  - items
 properties:
   id:
     description: The Subscription identifier string
     readOnly: true
     allOf:
       - $ref: "#/definitions/ResourceId"
+  orderType:
+    description: |
+      Specifies the type of order, will the customer receive the order by subscription
+      or is it a one-time purchase
+    type: string
+    enum:
+      - "subscription-order"
+      - "one-time-order"
+    default: "subscription-order"
+  status:
+    description: One-time order status
+    type: string
+    readOnly: true
   customerId:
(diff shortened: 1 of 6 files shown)


spec/definitions/Subscription.yaml
```diff
@@ -1,18 +1,41 @@
...
+    description: |
+      Specifies the type of order, will the customer receive the order by subscription
+      or is it a one-time purchase
+    type: string
+    enum:
+      - "subscription-order"
+      - "one-time-order"
+    default: "subscription-order"
+  status:
+    description: One-time order status
```
ashkarpetin: Is it only for one-time order?

slavcodev: Oops, copy-paste. The proper description is taken from `subscription-order.yaml` and `one-time-order`. The property here is added to manage the order of properties.

slavcodev:
```suggestion
```

spec/definitions/Subscription.yaml
```diff
@@ -1,18 +1,40 @@
...
+  - items
 properties:
   id:
     description: The Subscription identifier string
     readOnly: true
     allOf:
       - $ref: "#/definitions/ResourceId"
+  orderType:
+    description: |
+      Specifies the type of order, will the customer receive the order by subscription
```
adamaltman:
```suggestion
      Specifies the type of order, a subscription or a one-time purchase.
```

spec/definitions/Subscription.yaml
```diff
@@ -1,18 +1,40 @@
...
+      or is it a one-time purchase
```
adamaltman:
```suggestion
      
```
remove line

PR #542, opened 2018-11-28
Add bump offer to Transaction

colinbird:


spec/definitions/Transactions/Transaction.yaml
```diff
@@ -233,6 +233,61 @@ properties:
           - rejected
           - selected
           - unknown
+  hasBonus:
```
dara123: should we name this `hasBumpOffer` ?

spec/definitions/Transactions/Transaction.yaml
```diff
@@ -233,6 +233,61 @@ properties:
...
+    description: True if transaction has a bonus offer
+    type: boolean
+    readOnly: true
+  bonus:
```
dara123: `bumpOffer` ?

PR #544, opened 2018-12-05
Add tags endpoints

vova07:


spec/paths/tags.yaml
```diff
@@ -0,0 +1,59 @@
...
+          $ref: "#/headers/Rate-Limit-Limit"
+        Rate-Limit-Remaining:
+          $ref: "#/headers/Rate-Limit-Remaining"
+        Rate-Limit-Reset:
+          $ref: "#/headers/Rate-Limit-Reset"
+      schema:
+        $ref: "#/definitions/Tag"
+    401:
+      $ref: "#/responses/AccessForbidden"
+    409:
```
vova07: I added `409` code, for cases when someone try to create a tag with the name that already exists. 
Is it right, or should we allow creating multiple tags with the same name?

spec/paths/tags@{id}@customers.yaml
```diff
@@ -0,0 +1,73 @@
+parameters:
+- $ref: "#/parameters/resourceId"
+post:
```
vova07: I'm still not sure if we need an endpoint for multiple customers tagging and untagging.
Should I leave it?

vova07:
please review

spec/definitions/Tags/Tag.yaml
```diff
@@ -0,0 +1,20 @@
+type: object
+description: Tag
+required:
+- name
+properties:
+  id:
```
slavcodev: Is not the name unique? Why we can want another identifier?

vova07: I actually provide such question above.
Btw: tag is not a single word, so it may require some escapes when we'll use it in URL.

spec/paths/tags.yaml
adamaltman: It is right... the name should be unique.

spec/definitions/Tags/Tag.yaml
vova07: Your example show that we need an `edit` endpoint too.
I'll add it.

spec/definitions/Tags/Tag.yaml
```diff
@@ -0,0 +1,20 @@
...
+    description: The tag's identifier string
+    readOnly: true
+    allOf:
+    - $ref: "#/definitions/ResourceId"
+  name:
+    description: The tag's name
```
adamaltman:
```suggestion
    description: The tag's name, which must be unique.
```

Shouldn't there be a max length here?

spec/definitions/Tags/Tag.yaml
slavcodev: > I think because a tag isn't constrained by our url pattern -- but maybe it could be?
I do not know what `tags` we wish, but everywhere I have seen tags use very limited char set (github labels, gmail labels, etc). I think they all are safe to use in URL path.

spec/definitions/Tags/Tag.yaml
slavcodev: What about allowed pattern?

spec/definitions/Tags/Tag.yaml
```diff
@@ -11,6 +11,7 @@ properties:
   name:
     description: The tag's name, which must be unique.
     type: string
+    pattern: "[@~\-\.\w\d]+"
```
slavcodev: what about space?

vova07: Added `space`. 
`_` is included in `\w`.

spec/definitions/Tags/Tag.yaml
```diff
@@ -0,0 +1,31 @@
...
+  id:
+    description: The tag's identifier string
+    readOnly: true
+    allOf:
+    - $ref: "#/definitions/ResourceId"
+  name:
+    description: The tag's unique name, which is case-insensitive.
+    type: string
+    maxLength: 255
+    pattern: "^[@~\\-\\.\\w\\d ]+$"
```
adamaltman:
```suggestion
    pattern: "^[@~\\-\\.\\w\\d\\s]+$"
```

adamaltman: This is more clear and modern for `space`.

spec/paths/tags@{tag}@customers.yaml
```diff
@@ -0,0 +1,74 @@
+parameters:
+- $ref: "#/parameters/tag"
+post:
+  tags:
+  - Tags
+  summary: Tag a list of customers
+  description: |
+    Tag a list of customers
```
adamaltman: There are 3 customers with id's: `Adam`, `Brenda`, `Carl`.

There is one tag with id: `plays-soccer`.

`Carl` already has the tag:  `plays-soccer`.

What happens when you tag a list of customers and it includes Carl who already has the tag?

```
POST /tags/plays-soccer/customers
{
"customerIds": [
   "Adam",
   "Brenda",
   "Carl"
]
}
```
Will Carl end up with the same tag twice?  
Will it have no effect for Carl?  
Will there be an error?

My expectation is there is no effect for Carl. (I think)

It's just undocumented here, and if I was consuming the API, I wonder if I need to validate that first none of the customer id's I'm sending already have the tag.  If I needed to do that, would I have to do that one-by-one?

spec/swagger.yaml
```diff
@@ -388,6 +389,20 @@ parameters:
...
+    description: The customer resource identifier string
+    type: string
+    pattern: "^[@~\\-\\.\\w\\d]+$"
+    required: true
+  tag:
+    name: tag
+    in: path
+    description: The tag name
+    type: string
+    pattern: "^[@~\\-\\.\\w\\d ]+$"
```
adamaltman:
```suggestion
    pattern: "^[@~\\-\\.\\w\\d\s]+$"
```

spec/paths/tags@{tag}@customers.yaml
vova07: I think we can go with second option:
> Will it have no effect for Carl?
I do not have any objections here.
It seems to be easier to consume such API than doing additional validation before this call.

spec/paths/tags@{tag}@customers.yaml
```diff
@@ -0,0 +1,74 @@
...
+      - customerIds
+      properties:
+        customerIds:
+          description: The list of customer IDs
+          type: array
+          minItems: 1
+          items:
+            type: string
+  responses:
+    204:
```
vova07: I actually wonder if `204` is the right response code for this endpoint.
The batch insert as far as delete will be an asynchronous process.
So should we return `202` instead, or it's fine to leave it as is?

vova07:
please re-review

spec/paths/tags@{tag}@customers.yaml
adamaltman: I think it's fine to leave it as-is, but we should mention in the description that this is asynchronous and may therefor not be instantaneous.

spec/paths/tags@{tag}@customers.yaml
```diff
@@ -0,0 +1,58 @@
+parameters:
+- $ref: "#/components/parameters/tag"
+post:
```
adamaltman: This is `post` here, but in the endpoint below it is `put`.. is there a mistake in one or the other? why the inconsistency?

vova07: I may be wrong, but the decision was made because of the URL structure.
In the second case we know so far the `tag` and the `customerId` at the same time, so for me it was looking like a entity creation with predefined set of data.
Do you think it should it be changed?

adamaltman: Yes, it is inconsistent.

vova07: @adamaltman can you please advice me which method it's better to use for both endpoints?
For me it looks natural and more logical as it is now.
But if you think it's confusing, I'll be glad to change it, but just need an advice with which method to do it.

spec/paths/tags@{tag}@customers@{customerId}.yaml
```diff
@@ -0,0 +1,29 @@
+parameters:
+- $ref: "#/components/parameters/tag"
+- $ref: "#/components/parameters/customerId"
+put:
```
adamaltman: Change this one to post.

vova07:
please re-review

PR #545, opened 2018-12-06
Add billing timing for Plan pricing

slavcodev:


slavcodev:
devs please re-review

spec/definitions/Plans/PlanBillingTiming.yaml
```diff
@@ -0,0 +1,9 @@
+type: string
+default: "prepaid"
+description: |
+  The billing timing in relation to the service period.
+  For `prepaid` plans the customer pays when service period starts,
+  whereas for `postpaid` he pays when service period ends.
+enum:
+  - "prepaid"
+  - "postpaid"
```
2e3s: Can there be ever other values that those 2? Can it be just `isPostpaid` boolean?

slavcodev: It can be. By why not `isPrepaid`? I'd prefer string value rather than boolean flag i this case.

slavcodev: > Can there be ever other values that those 2?
May be `random` :) Or `auto`

2e3s: > I'd prefer string value rather than boolean flag i this case.

Well, every boolean API can be replaced by 2 strings value, but yet. It just seemed logical to me, if ever use boolean values.
`random` and `auto` I think can be applied to any value too)

2e3s: the benefit of a boolean instead of 2-values strings is that when you see on a boolean you know immediately the range of possible values. With strings, you have to learn API

slavcodev: Alright, last comment was just a joke. But more seriously, it can be `null` too. In this case we use default value to be back compatible with existing integration. But in other places, e.q. subscription it would be nullable with default to null. And I know that flag property `isPostpaid` also can be nullable, but it already not so "logical": `isPostpaid: null`.

slavcodev: > With strings, you have to learn API

You have to learn API anyway, don't you? How you will know about existence of property and its format (bool)? With string values there is not more complicated, it's the same you need to learn the format and specs tells explicitly the `enum`.

slavcodev: In other hand, the benefit of strings, that doing integration, devs know that this property is string, you can't rely on its value and should check it, etc. With boolean property the checking is different.
This gives us more opportunities to evolve our API.

spec/definitions/Plans/PlanBillingTiming.yaml
```diff
@@ -0,0 +1,9 @@
+type: string
+default: "prepaid"
+description: |
+  The billing timing in relation to the service period.
+  For `prepaid` plans the customer pays when service period starts,
```
adamaltman:
```suggestion
  For `prepaid` plans the customer pays when the service period starts,
```

adamaltman:
```suggestion
  For `prepaid` plans the customer pays when service period starts;
```

spec/definitions/Plans/PlanBillingTiming.yaml
```diff
@@ -0,0 +1,9 @@
...
+  whereas for `postpaid` he pays when service period ends.
```
adamaltman:
```suggestion
  whereas, for `postpaid` plans, the customer pays when the service period ends.
```

spec/definitions/Plans/PlanBillingTiming.yaml
adamaltman: Sometimes it's hard to envision a non-binary outcome.  What if there is a trend in the for `midterm` payments?  With this design we can simply add a `midterm` option.

PR #559, opened 2018-12-19
Specify subscription statuses

2e3s:
At the moment outdated values are displayed: 

- `Active`
- `Will become active at a future date`
- `Active but set to cancel at next rebill date`
- `Cancelled`
- `Inactive`

### DIFF
--- a/spec/components/schemas/Subscription.yaml
+++ b/spec/components/schemas/Subscription.yaml
@@ -24,7 +24,7 @@ properties:
     type: string
     readOnly: true
     enum:
-      - draft
+      - pending
       - active
       - canceled
       - churned
--- a/spec/components/schemas/Subscription.yaml
+++ b/spec/components/schemas/Subscription.yaml
@@ -23,6 +23,12 @@ properties:
   status:
     type: string
     readOnly: true
+    enum:
+      - draft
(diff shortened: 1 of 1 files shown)


spec/components/schemas/Subscription.yaml
```diff
@@ -23,6 +23,12 @@ properties:
   status:
     type: string
     readOnly: true
+    enum:
+      - draft
```
slavcodev: Just wonder, might the `pending` be alternative variant?

2e3s: This state will be applicable when the subscription doesn't start (not paid etc). Would it stay forever in `pending` state? I think it would be confusing.

slavcodev: What do you mean? The subscription is not considered `draft` if its `start time` passed, even invoice is not paid.

2e3s: You mentioned once that we're going to get rid of cancelled never-started so that it would just stay draft. This is what I mean.

2e3s: Please, expand your idea if I'm wrong

PR #561, opened 2018-12-20
Improve payment instruments payload to make it clear which method

adamaltman:


XaosSintez:
I'm not sure i understand this.
These schemas are used in their own endpoints, like this:
https://github.com/Rebilly/RebillyAPI/blob/89f5a9e29ea8c223d1ecef67d7ac2ad79863543d/spec/paths/payment-cards@%7Bid%7D.yaml#L113

Is `method` property read only or client needs to set it? Because client can send request to `PUT /payment-cards/{id}` with `method: "paypal"`

PR #567, opened 2018-12-28
Minor fixes to the documentation

adamaltman:


spec/paths/bank-accounts.yaml
```diff
@@ -1,3 +1,33 @@
...
+        Rate-Limit-Reset:
+          $ref: "#/components/headers/Rate-Limit-Reset"
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/BankAccount"
+    401:
+      $ref: "#/components/responses/AccessForbidden"
+    422:
+      description: Invalid data was sent
```
ashkarpetin: `$ref: "#/components/responses/InvalidDataError"`

spec/paths/contacts.yaml
```diff
@@ -36,7 +38,14 @@ post:
     - Contacts
   summary: Create a contact
   description: |
-    Create a contact
+    Contacts are created automatically by data passed to the 
+    contact-related hashes within the customer, invoice,
+    payment instrument (like payment card), subscriptions,
+    and more.    
+    <div class="alert alert-warning">
```
ashkarpetin: Unnecessary html tags

adamaltman: It is required, afaiu, there is no markdown tag for warning.

creocoder: @ashkarpetin How you will mark text inside that block as alert?

spec/paths/bank-accounts.yaml
creocoder: @ashkarpetin Good catch, but in that case all other pathes should be refactored same way. For example:

https://github.com/Rebilly/RebillyAPI/blob/master/spec/paths/subscriptions.yaml#L69

spec/paths/contacts.yaml
ashkarpetin: I didn't know about that feature..

PR #568, opened 2018-12-28
Improve errors

adamaltman:
Use the appropriate error reference.

Augment the error description to match our current responses.

(Note: I accidentally included the package lock file... prevented in the future by adding it to the .gitignore)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,4 +33,4 @@ node_modules
 .node_repl_history
 
 # npm
-pacakge-lock.json
+package-lock.json
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@ node_modules
 
 # Optional REPL history
 .node_repl_history
+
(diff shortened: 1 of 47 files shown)


adamaltman:
I did not mean to include package-lock.json

.gitignore
```diff
@@ -31,3 +31,6 @@ node_modules
 
 # Optional REPL history
 .node_repl_history
+
+# npm
+pacakge-lock.json
```
2e3s: looks wrong

adamaltman: what looks wrong about it?

2e3s: `pacakge` should be `package`

adamaltman:
```suggestion
package-lock.json
```

adamaltman: Thanks, I resolved it.

PR #570, opened 2018-12-30
Changed payments request and response to be in line with transactions

adamaltman:


spec/components/schemas/PaymentInstrument.yaml
```diff
@@ -6,3 +6,12 @@ discriminator:
 properties:
   method:
     $ref: "#/components/schemas/Method"
+  gatewayAccountId:
```
adamaltman: Due to other discussions, we'll remove this here.

adamaltman:
Made #572 instead.

PR #572, opened 2019-01-03
Add Transaction Request API

adamaltman:


spec/components/schemas/Transactions/TransactionRequest.yaml
```diff
@@ -0,0 +1,65 @@
...
+  type:
+    description: |
+      The type of transaction requested. 
+      You should always include the type within your API request.
+      This supports a limited subset of Transaction types.  To refund or void, use the refund endpoint.
+      To `capture` use the `sale` endpoint.  It will capture if there are any eligible
+      transactions of type `authorize` that can be captured.
+    type: string
+    enum:
+    - sale
```
ivizma: please fix alignment

adamaltman: I fixed it.  Thank you.

spec/components/schemas/Transactions/TransactionRequest.yaml
```diff
@@ -0,0 +1,65 @@
+type: object
+required:
+  - websiteId
+  - customerId
+  - currency
+  - amount
+  - type
+properties:
```
dara123: should we have `redirectUrl` here too?

adamaltman: I don't think so... we can discuss it more.

dara123:
need to merge this for https://github.com/Rebilly/backend/pull/3909 test

PR #575, opened 2019-01-09
Replace invoice isRefunded and isDisputed with appropriate statuses

2e3s:
As I understand, an invoice can be disputed only when it's paid, and paid status cannot change, so this would be logical to make as states

### DIFF
--- a/spec/components/schemas/Invoice.yaml
+++ b/spec/components/schemas/Invoice.yaml
@@ -96,14 +96,8 @@ properties:
      - "paid"
      - "abandoned"
      - "voided"
-  isDisputed:
-    type: string
-    description: Whether the invoice (related transaction) has been disputed
-    readOnly: true
-  isRefunded:
-    type: string
-    description: Whether the invoice (related transaction) has been subsequently refunded
-    readOnly: true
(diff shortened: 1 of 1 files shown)


spec/components/schemas/Invoice.yaml
```diff
@@ -96,14 +96,8 @@ properties:
...
      - "voided"
-  isDisputed:
-    type: string
-    description: Whether the invoice (related transaction) has been disputed
-    readOnly: true
-  isRefunded:
-    type: string
-    description: Whether the invoice (related transaction) has been subsequently refunded
-    readOnly: true
+     - "disputed"
```
dara123: I think invoice can be both `paid` and `disputed`

2e3s: It's probably like "invoice can be both draft and issued". I think one always follow the other.

2e3s: I mean, invoice cannot be both paid + disputed and non-paid + disputed

adamaltman: I think this is good.

PR #576, opened 2019-01-10
Added KYC purpose

adamaltman:


destebang:
This has been almost a year in PR, @adamaltman  does this PR still apply?

adamaltman:
Yes, it does.  Hard to get a review around here. ;-)

spec/components/schemas/KycDocument.yaml
```diff
@@ -23,6 +23,11 @@ properties:
     enum:
       - identity-proof
       - address-proof
+  purpose:
```
destebang: Would it make sense to restrict this field to a nullable `enum` ?

I can not imagine too many reasons to request a KYC check (basically it can serve the purpose of accepting a customer or to send money to someone). Having an enum would also allow us to get an understanding of what is the most common use of this tool.

Also, we always can evolve from an enum to a string while the opposite would be more difficult. 
I am not sure enough about all the possible values the enum could accept, though.

destebang: Those values seem ok to me as a starting point.

mclassic: @adamaltman Those possibilities seem okay for now. We can always add to or change the enum if needed.

PR #579, opened 2019-01-11
Added readOnly `isUsed` property to PaymentToken schema

justnero-ru:


spec/components/schemas/PaymentToken.yaml
```diff
@@ -7,6 +7,11 @@ properties:
     readOnly: true
     allOf:
       - $ref: "#/components/schemas/ResourceId"
+  used:
```
ashkarpetin: We use `is` prefix for boolean fields/props. I think `isUsed` or `isApplied` wil be better.

PR #588, opened 2019-01-15
Add fingerprint to bank account resource

ashkarpetin:


spec/components/schemas/BankAccount.yaml
```diff
@@ -46,6 +46,10 @@ properties:
     description: The Address. Required if bank account is not created from Token
     allOf:
       - $ref: "#/components/schemas/ContactObject"
+  fingerprint:
+    description: A unique value to identify the Bank Account.  It contains alphanumeric values.
```
wmdmgv: `Account.  It`
two space chars?

ashkarpetin: Good question. There are many places with two space chars e.g. https://github.com/Rebilly/RebillyAPI/pull/572. Can't figure out is it intentionally or not.

PR #589, opened 2019-01-15
Add write-only Subscription.couponIds

2e3s:
This is write-only for now, the functionality may be extended to be read later.

### DIFF
--- a/spec/components/schemas/Subscription.yaml
+++ b/spec/components/schemas/Subscription.yaml
@@ -102,3 +102,10 @@ properties:
     description: Subscription activation time
     allOf:
       - $ref: "#/components/schemas/ServerTimestamp"
+  couponIds:
+    type: array
+    description: A list of coupons to redeem on the customer and restrict by this subscription
+    writeOnly: true
+    items:
+      type: string
+      description: Coupon ID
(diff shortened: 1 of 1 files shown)


spec/components/schemas/Subscription.yaml
```diff
@@ -102,3 +102,10 @@ properties:
     description: Subscription activation time
     allOf:
       - $ref: "#/components/schemas/ServerTimestamp"
+  couponIds:
+    type: array
+    description: A list of coupons to redeem on the customer and restrict by this subscription
```
adamaltman:
```suggestion
    description: A list of coupons to redeem on the customer and restrict by this subscription
      If the customer already has redeemed the coupon, it will have no effect.  
```

Is my suggestion correct?

2e3s: No, not really. A customer can redeem a coupon multiple times (with different restrictions). So I'm doing an unconditional redemption with the mentioned restriction.

adamaltman:
```suggestion
    description: A list of coupons to redeem on the customer and restrict to this subscription.
      Read more about [coupons here](https://help.rebilly.com/invoices-and-subscriptions/coupons-discounts).
```
The restriction wasn't clear to me.

PR #600, opened 2019-01-18
Improve the getting started docs

adamaltman:


web/description.md
```diff
@@ -346,9 +346,14 @@ PUT customers/<id>
 ```
 
 Contained within the customer creation response is
-a `customerId` and a `defaultPaymentInstrument`.
+an `id`.  The `id` is utilized within other
+operations as the `customerId`.  It will match
+the id value used in the path paramter of the `PUT`
```
ashkarpetin:
```suggestion
the id value used in the path parameter of the `PUT`
```

web/description.md
```diff
@@ -482,4 +487,18 @@ If the payment result is `declined` you may
 present a message to the customer to retry 
 with different payment information.
 
+A sample request:
+```
+POST transactions
+{
+    "customerId": "<my customer id used above>",
+    "type": "sale",
+    "orderId": "<generate a unique identifier, such as a UUID>",
```
adamaltman: I find it strange this is called `orderId` and thing we should name it differently (it's not yet in the API spec)... it could be a `requestId` for example.

2e3s: What is the reason this value is here?

PR #601, opened 2019-01-18
Add request id and billing descriptor

adamaltman:


colinbird:
Can we change `orderId` to `requestId` here ? https://github.com/Rebilly/RebillyAPI/blob/967e22878ef2f0ef28a37cbb39944c8001c30161/spec/components/schemas/Transactions/Transaction.yaml#L360

adamaltman:
Yes, I think we can...  but instead, we may want to add requestId and deprecate orderId.  And then remove orderId after confirming no usages of it.

PR #606, opened 2019-01-22
Remove token expiration endpoint mentioning from description

ashkarpetin:
Endpoint was removed in #585


ashkarpetin:
devs please review

PR #610, opened 2019-01-24
Changed subscription and invoice statuses

adamaltman:
- added Subscription billingStatus

### DIFF
--- a/spec/components/schemas/Subscription.yaml
+++ b/spec/components/schemas/Subscription.yaml
@@ -22,13 +22,34 @@ properties:
     default: "subscription-order"
   status:
     type: string
+    description: |
+      The status of the subscription service. A subscription starts
+      in the `pending` status, and will become `active` when the 
+      service period begins.
     readOnly: true
     enum:
       - pending
       - active
       - canceled
       - churned
+      - suspended
       - paused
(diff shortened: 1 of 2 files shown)


spec/components/schemas/Subscription.yaml
```diff
@@ -22,13 +22,34 @@ properties:
...
+      - suspended
       - paused
+  billingStatus:
+    description: |
+      The billing status of the most recent invoice.  It may
+      help you determine if you should change the service status
+      such as suspending the service.
+    type: string
+    readOnly: true
+    enum:
```
slavcodev: Maybe would be nice to segregate it into new definition in order to re-use.

vova07: Yes, so we should not update 2 places in case it will change in the future.

2e3s: It's probably true, probably not worth it: not all invoice statuses may be supported here like `draft`

adamaltman: That's why I didn't do that (what @2e3s said)

slavcodev: Which billing status would have the draft subscription?

PR #612, opened 2019-01-25
Rename redemptionCode to id/couponId

2e3s:
Coupon is the only resource which has its identifier named not as `id` but `redemptionCode`. At the same time there are redemptions. This whole situation seems confusing to me.


2e3s:
No, it has never been approved

PR #615, opened 2019-01-29
Add missing event types to Order Timeline

XaosSintez:


spec/components/schemas/Timelines/OrderTimeline.yaml
```diff
@@ -26,6 +26,11 @@ properties:
       - invoice-partially-paid
       - subscription-upgraded
       - subscription-downgraded
+      - subscription-billing-address-changed
```
adamaltman: I think we should be naming these things `order` instead of `subscription`.... Perhaps that's a different project.

PR #631, opened 2019-02-11
Rename Order Timeline events

XaosSintez:


edtoken:
awaiting front-end update [GWY-10732](https://jira.ilabsinc.com/browse/GWY-10732)
in progress

PR #632, opened 2019-02-11
Add pendingReason to Subscription

2e3s:


2e3s:
devs please review

spec/components/schemas/Subscription.yaml
```diff
@@ -34,6 +32,23 @@ properties:
...
       - suspended
       - paused
+  pendingReasons:
+    type: array
+    description: |
+      A list of reasons why the subscription should stay in the pending status.
+      - "initial-payment" blocks the activation until the initial invoice is paid
+      - "start-time" blocks the activation until startTime < now
+      - "custom" is to be used when a business requires a later activation.
+        The subscription won't be activated until the field is updated without this pending reason.
```
adamaltman: I think we can explain this more clearly.  I think it's the right idea.

2e3s:
@slavcodev has suggested another approach which I've summarized in #641
@adamaltman please, review this proposal too.

2e3s:
It has become outdated, the solution should be redefined.

PR #641, opened 2019-02-20
Add an ability to make subscription pending or active

2e3s:
Alternative approach suggested by @slavcodev from https://github.com/Rebilly/RebillyAPI/pull/632

### DIFF
--- a/spec/components/schemas/Subscription.yaml
+++ b/spec/components/schemas/Subscription.yaml
@@ -35,7 +35,7 @@ properties:
       - paused
   autoActivation:
     type: boolean
-    descrition: |
+    description: |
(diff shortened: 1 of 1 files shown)


spec/components/schemas/Subscription.yaml
```diff
@@ -23,17 +23,22 @@ properties:
   status:
     type: string
     description: |
-      The status of the subscription service. A subscription starts
-      in the `pending` status, and will become `active` when the 
-      service period begins.
-    readOnly: true
```
2e3s: I don't like that here would be a special implicit action "activate" on changing the status

2e3s:
It has become outdated, the solution should be redefined.

PR #651, opened 2019-03-06
Add missing order events to Customer timeline

viceromag:


spec/components/schemas/Timelines/CustomerTimeline.yaml
```diff
@@ -51,6 +51,9 @@ properties:
       - transaction-amount-discrepancy-found
       - email-message-sent
       - custom-event-processed
+      - order-upgraded
+      - order-downgraded
+      - order-churned
```
ashkarpetin: Probably would be better to group with other order events above

PR #653, opened 2019-03-06
Add invoice autopay attributes

vova07:


vova07:
please review

spec/components/schemas/Invoices/Invoice.yaml
```diff
@@ -90,6 +90,17 @@ properties:
     items:
       allOf:
         - $ref: "#/components/schemas/InvoiceDiscount"
+  autopayScheduledTime:
+    description: Invoice autopay scheduled time
+    readOnly: true
```
adamaltman: This should not be read only.

PR #654, opened 2019-03-11
Adjust autopayScheduledTime attribute

vova07:


spec/components/schemas/Invoices/Invoice.yaml
```diff
@@ -92,7 +92,7 @@ properties:
         - $ref: "#/components/schemas/InvoiceDiscount"
   autopayScheduledTime:
     description: Invoice autopay scheduled time
-    readOnly: true
+    readOnly: false
```
ashkarpetin:
```suggestion
```

vova07: @ashkarpetin I need to make it explicitly `false` to override the value from `ServerTimestamp`

ashkarpetin: Then you are using wrong schema:
```
type: string
description: Read-only timestamp, automatically assigned on back-end.
format: date-time
readOnly: true
```

spec/components/schemas/Invoices/Invoice.yaml
```diff
@@ -92,7 +92,7 @@ properties:
...
     allOf:
```
ashkarpetin:
```suggestion
```

spec/components/schemas/Invoices/Invoice.yaml
ashkarpetin: I suggest add this and remove ` $ref: "#/components/schemas/ServerTimestamp"`

slavcodev: Yes, the purpose of `ServerTimestamp` schema is the read-only time which is set by server. For writable fields, no reasons to use this schema.

PR #659, opened 2019-03-21
Add merge customers endpoint

2e3s:


spec/components/schemas/CustomerMerge.yaml
```diff
@@ -0,0 +1,5 @@
+type: array
+description: List of changes for the merge
```
2e3s: This is not trivial nor stable to implement. I'm not sure about the value of this. If there would be no response or just a redirect to the target customer it would be much simpler.

spec/paths/customers@{duplicateId}@merge-to@{targetId}.yaml
```diff
@@ -0,0 +1,38 @@
...
+      type: string
+  - name: targetId
+    in: path
+    description: The target customer's ID
+    required: true
+    schema:
+      type: string
+post:
+  tags:
+    - Subscriptions
```
VladimirBerdnik: Is this tag correct? Shouldn't it be `Customers`?

spec/paths/customers@{duplicateId}@merge-to@{targetId}.yaml
```diff
@@ -0,0 +1,38 @@
...
+        Rate-Limit-Remaining:
+          $ref: "#/components/headers/Rate-Limit-Remaining"
+        Rate-Limit-Reset:
+          $ref: "#/components/headers/Rate-Limit-Reset"
+      content:
+        application/json:
+          schema:
+            type: array
+            items:
+              $ref: "#/components/schemas/CustomerMerge"
```
adamaltman: Yes, I disagree with this content here.  It's unclear and probably unusable by a client too. I'd recommend 204 no content.

spec/paths/customers@{duplicateId}@merge-to@{targetId}.yaml
```diff
@@ -0,0 +1,38 @@
...
+    required: true
+    schema:
+      type: string
+post:
+  tags:
+    - Customers
+  summary: Merge one customer to another one
+  description: Merge one duplicate customer to another target customer and delete the former.
+  responses:
+    200:
```
adamaltman: I'd just this to 204 no content.

spec/components/schemas/CustomerMerge.yaml
adamaltman: I agree -- I would remove this, and do a 204 no content.

spec/paths/customers@{duplicateId}@merge-to@{targetId}.yaml
```diff
@@ -0,0 +1,32 @@
...
+    required: true
+    schema:
+      type: string
+  - name: targetId
+    in: path
+    description: The target customer's ID
+    required: true
+    schema:
+      type: string
+post:
```
2e3s: I suspect that this has to be `delete`.
It's semantically the same as `customers@{duplicateId}` but it has additional parameters.

2e3s:
devs please review

spec/paths/customers@{id}.yaml
```diff
@@ -46,3 +46,29 @@ put:
       $ref: "#/components/responses/Conflict"
     422:
       $ref: "#/components/responses/InvalidDataError"
+delete:
```
adamaltman: I don't think it should be delete, because it doesn't really delete.  It merges it into another customer.

2e3s: So in another place we agreed on that it looks like a delete

PR #660, opened 2019-03-25
Add upcoming invoices endpoints

2e3s:
These are not permanent and will be removed after full integration.

### DIFF
--- a/spec/paths/customers@{id}@upcoming-invoices.yaml
+++ b/spec/paths/customers@{id}@upcoming-invoices.yaml
@@ -1,0 +1,31 @@
+parameters:
+  - $ref: "#/components/parameters/resourceId"
+get:
+  tags:
+    - Invoices
+  summary: Retrieve customer's upcoming invoices
+  description: |
+    Retrieve a list of upcoming invoices from the subscriptions which belong to the given customers.
+    The endpoint is temporary before upcoming invoices get a complete integration.
+  parameters:
+    - $ref: "#/components/parameters/collectionExpand"
+  responses:
+    200:
+      description: Upcoming invoices are retrieved successfully.
+      headers:
+        Rate-Limit-Limit:
+          $ref: "#/components/headers/Rate-Limit-Limit"
+        Rate-Limit-Remaining:
+          $ref: "#/components/headers/Rate-Limit-Remaining"
+        Rate-Limit-Reset:
+          $ref: "#/components/headers/Rate-Limit-Reset"
+      content:
+        application/json:
+          schema:
+            type: array
+            items:
+              $ref: "#/components/schemas/Invoice"
+    401:
+      $ref: "#/components/responses/AccessForbidden"
+    404:
+      $ref: "#/components/responses/NotFound"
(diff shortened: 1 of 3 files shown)


spec/paths/subscriptions@{id}@upcoming-invoice.yaml
```diff
@@ -0,0 +1,29 @@
...
+        Rate-Limit-Limit:
+          $ref: "#/components/headers/Rate-Limit-Limit"
+        Rate-Limit-Remaining:
+          $ref: "#/components/headers/Rate-Limit-Remaining"
+        Rate-Limit-Reset:
+          $ref: "#/components/headers/Rate-Limit-Reset"
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/Invoice"
```
2e3s: This endpoint is only for 1 invoice. Not sure if many is needed, it would just formal as now multiple upcoming invoices is not possible (and may not be possible in the future if designed so).

spec/paths/subscriptions@{id}@upcoming-invoice.yaml
```diff
@@ -0,0 +1,29 @@
+parameters:
+  - $ref: "#/components/parameters/resourceId"
+get:
+  tags:
+    - Invoices
+  summary: Retrieve subscription's upcoming invoice
+  description: |
+    Retrieve an upcoming invoice from the specified subscriptions.
```
ashkarpetin:
```suggestion
    Retrieve an upcoming invoice from the specified subscription.
```

spec/paths/subscriptions@{id}@upcoming-invoice.yaml
2e3s: On the other hand, since this is an experimental endpoint, it could be many invoices in future and it's easier to be prepared for that.

spec/paths/customers@{id}@upcoming-invoices.yaml
```diff
@@ -0,0 +1,31 @@
+parameters:
+  - $ref: "#/components/parameters/resourceId"
+get:
+  tags:
+    - Invoices
+  summary: Retrieve customer's upcoming invoices
+  description: |
+    Retrieve a list of upcoming invoices from the subscriptions which belong to the given customers.
```
XaosSintez:
```suggestion
    Retrieve a list of upcoming invoices from the subscriptions which belong to the given customer.
```

spec/paths/subscriptions@{id}@upcoming-invoices.yaml
```diff
@@ -0,0 +1,31 @@
+parameters:
+  - $ref: "#/components/parameters/resourceId"
+get:
+  tags:
+    - Invoices
+  summary: Retrieve subscription's upcoming invoice
+  description: |
+    Retrieve an upcoming invoice from the specified subscriptions.
```
XaosSintez:
```suggestion
    Retrieve an upcoming invoice from the specified subscription.
```

PR #661, opened 2019-03-25
Customer authentication

justnero-ru:
Expose AuthenticationToken endpoints for publishable key auth and add token exchange for JWT session endpoint


spec/paths/authentication-tokens@{token}.yaml
```diff
@@ -29,12 +33,47 @@ get:
...
     404:
       $ref: "#/components/responses/NotFound"
+post:
+  tags:
+    - Customer Authentication
+  summary: Exchange
+  description: |
+    Exchange Authentication Token for JWT
+
+    It will also logout a used under this token
```
adamaltman: A used what?

adamaltman: Or a user?

justnero-ru: typo fixed

spec/paths/authentication-tokens@{token}.yaml
```diff
@@ -29,12 +33,47 @@ get:
...
     404:
       $ref: "#/components/responses/NotFound"
+post:
+  tags:
+    - Customer Authentication
+  summary: Exchange
+  description: |
+    Exchange Authentication Token for JWT
+
+    It will also logout a user under this token
```
adamaltman: I wonder what this means.  By exchanging an authentication token for a JWT, you logout any already logged-in sessions for the same token? Or what exactly?

justnero-ru: It should mean we remove this token after exchanging, so it will be no longer valid and can be used just once.

justnero-ru:
devs please re-review

spec/paths/authentication-tokens@{token}.yaml
```diff
@@ -29,15 +33,50 @@ get:
...
     404:
       $ref: "#/components/responses/NotFound"
+post:
+  tags:
+    - Customer Authentication
+  summary: Exchange
+  description: |
+    Exchange Authentication Token for JWT
+
+    It will also make Authentication Token non-valid any more, so it can be exchanged only once.
```
adamaltman:
```suggestion
    It will also invalidate an Authentication Token (so it can be exchanged only once).
```

justnero-ru:
devs please review

spec/paths/authentication-tokens@{token}.yaml
```diff
@@ -29,15 +33,50 @@ get:
...
     404:
       $ref: "#/components/responses/NotFound"
+post:
+  tags:
+    - Customer Authentication
+  summary: Exchange
+  description: |
+    Exchange Authentication Token for JWT
+
+    It will also invalidate an Authentication Token (so it can only be exchanged once).
```
adamaltman: I'm not sure this should happen automatically.  Maybe it should be a flag on the request or someone should make a subsequent request to invalidate the token.

justnero-ru: Added an option for that. By default it will invalidate token, but can be set to false for multiple use

PR #665, opened 2019-03-27
Add missing scheduling method

ashkarpetin:


spec/components/schemas/Scheduling/InvoiceRetryScheduleInstruction.yaml
```diff
@@ -6,6 +6,7 @@ properties:
   method:
     type: string
     enum:
+      - "auto"
```
adamaltman: We do not want this.

spec/components/schemas/Scheduling/README.md
```diff
@@ -9,6 +9,7 @@ Strategies how to calculate relative time. It answers the question of "when" to
 Available methods:
 
 - auto
```
adamaltman: remove auto

PR #672, opened 2019-04-01
Add Gigadat related methods

XaosSintez:


spec/components/schemas/Method.yaml
```diff
@@ -25,6 +25,7 @@ enum:
   - EPS
   - ePay-bg
   - Flexepin
+  - Gigadat
```
adamaltman: This is not a method.

PR #675, opened 2019-04-04
Reformat of invalidFields in the error response

slavcodev:


spec/components/schemas/InvalidError.yaml
```diff
@@ -1,10 +1,21 @@
...
+        type: array
+        items:
+          type: object
+          properties:
+            field:
+              type: string
+              description: The field name. Dot notation is used for nested object field names.
+            message:
+              type: string
+        example:
```
2e3s: Don't we need `|` or `>` for multiline in yaml?

slavcodev: No. `|` or `>` is for strings only. The `example` key has mized format, in this case it's `json`

PR #682, opened 2019-04-19
Fix Customer Auth passwordless typo

justnero-ru:
Actually should be `passwordless` and not `password-less`. It was discussed earlier on the backend side

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,3 @@ node_modules
 
 # npm
 package-lock.json
-
-# JetBrains IDEs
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -34,3 +34,6 @@ node_modules
 
 # npm
 package-lock.json
+
+# JetBrains IDEs
+.idea
```
adamaltman: keep this as a global gitignore instead of adding to this package gitignore.

PR #688, opened 2019-05-09
Add transaction revision to API specs

mbezhanov:
Adds the transaction revision field to the API specs

### DIFF
--- a/spec/components/schemas/Transactions/Transaction.yaml
+++ b/spec/components/schemas/Transactions/Transaction.yaml
@@ -184,6 +184,7 @@ properties:
   revision:
     description: The number of times the transaction data has been modified
     type: integer
+    readOnly: true
   bin:
     description: Payment Card BIN
     type: string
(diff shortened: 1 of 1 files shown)


spec/components/schemas/Transactions/Transaction.yaml
```diff
@@ -181,6 +181,9 @@ properties:
   velocity:
     description: The number of transactions by the same customer in the past 24 hours
     type: integer
+  revision:
```
slavcodev: read-only rght?

spec/components/schemas/Transactions/Transaction.yaml
```diff
@@ -181,6 +181,9 @@ properties:
...
+    description: The number of times the transaction data has been modified
+    type: integer
```
slavcodev:
```suggestion
    type: integer
    readOnly: true
```


read-only rght?

PR #692, opened 2019-05-10
Update search filter description

mclassic:
Update the search filter description to explain `date` parameter.

### DIFF
--- a/spec/paths/search.yaml
+++ b/spec/paths/search.yaml
@@ -15,7 +15,8 @@ get:
             The collection items filter requires a special format.
             Use "," for multiple allowed values.  Use ";" for multiple fields.
             Use one or more of the listed filter fields. i.e. `is:customer email:bob@test.com`.
-            The `date` value can be surrounded with quotes and accepts [any format that PHP's `strtotime` supports.](https://www.php.net/manual/en/function.strtotime.php)
+            The `date` value can be surrounded with quotes. Examples:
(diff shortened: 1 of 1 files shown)


spec/paths/search.yaml
```diff
@@ -14,7 +14,8 @@ get:
         - description: |
             The collection items filter requires a special format.
             Use "," for multiple allowed values.  Use ";" for multiple fields.
-            Use one or more of the listed filter fields. i.e. `is:customer email:bob@test.com`
+            Use one or more of the listed filter fields. i.e. `is:customer email:bob@test.com`.
+            The `date` value can be surrounded with quotes and accepts [any format that PHP's `strtotime` supports.](https://www.php.net/manual/en/function.strtotime.php)
```
2e3s: Can't the API be less technology-dependent?

2e3s: It's better to describe the particular usage cases which we will certainly support whatever technology is used for this endpoint(s).

mclassic: It will make the description field very long. Can you suggest where else I could put it?

2e3s: How long? Do users need a zillion use cases, everyone with that std library function provides? I think putting here is okay.

mclassic: Okay, I'll add some and get rid of this link.

mclassic:
devs please review

PR #696, opened 2019-05-13
Adjust customer JWT schema

ashkarpetin:


ashkarpetin:
@slavcodev idk it's permissions prop is returned from api response

slavcodev:
I believe this attribute make non sense at the moment at least. There is no way to set the permissions to customer, isn't it? I think customer  has access to their data w/o restrictions and that's it.

PR #698, opened 2019-05-15
Add embedded property to specific resources

vova07:
## Description
Add `_embedded` property to specific resources.

## Note
I've start adding new property to all resources that support it, but after some backend tests, I think we should complete this step-by-step, as not all relations are supported on backend, because of that we may have a situation when F-E provide more conditions we support on backend, and it may be confusing.


spec/components/schemas/Criteria/Condition.yaml
```diff
@@ -0,0 +1,20 @@
+type: object
```
adamaltman: Where is this used within RebillyAPI?

adamaltman: I see because you added GatewayAccount here.  I believe instead you could include GatewayAccount by URL reference...

vova07: In the `GatewayAccount` schema which was added.
https://github.com/Rebilly/RebillyAPI/pull/698/files#diff-6f826096ab9bc301da86100c91f27caeR107

vova07:
please review

vova07:
please review

vova07:
please review

vova07:
please review

PR #710, opened 2019-06-12
Add kyc status to customer resource

ashkarpetin:


spec/components/schemas/Customer.yaml
```diff
@@ -50,6 +50,13 @@ properties:
     type: array
     items:
       $ref: "#/components/schemas/Tag"
+  kyc:
```
dara123: what if customer haven't upload kyc yet?

ashkarpetin: It would be null if customer haven't upload kyc or if it's status is not `accepted/rejected`

ashkarpetin: Probably better add this information to description. Also I'm not sure about the property name..

PR #711, opened 2019-06-14
Add referenceData to Transaction API

viceromag:


spec/components/schemas/Transactions/Transaction.yaml
```diff
@@ -186,6 +186,14 @@ properties:
     description: The number of times the transaction data has been modified
     type: integer
     readOnly: true
+  referenceData:
+    description: Transaction reference data
+    type: object
```
adamaltman: Add a new line:

`nullable: true`

PR #712, opened 2019-06-17
Make Subscription.orderType read-only

2e3s:


XaosSintez:
It will depend on items i guess? No way to set it manually?

2e3s:
Yes, exactly. It already depends on the items, if you post inconsistent items and orderType it will throw an error.

2e3s:
Closing for now, there is no good from this, I guess

PR #718, opened 2019-06-25
Customer Timeline improvements

XaosSintez:
This PR improves Customer Timeline::

1. Adds `CustomerTimelineCustomEvent` and corresponding `customer-timeline-custom-events` endpoint to add custom event types
2. Adds `custom-event` type to Customer Timeline and adds `customEventType`, `customData` fields
3. Adds `customer-timeline-events` endpoint to be able to get Customer Timeline events for all Customers

### DIFF
--- a/spec/components/schemas/Timelines/CustomerTimelineCustomEvent.yaml
+++ b/spec/components/schemas/Timelines/CustomerTimelineCustomEvent.yaml
@@ -7,6 +7,5 @@ properties:
     type: string
   createdTime:
     description: Customer Timeline Custom event created time
-    readOnly: true
     allOf:
       - $ref: "#/components/schemas/ServerTimestamp"
--- a/spec/components/schemas/Timelines/CustomerTimelineCustomEvent.yaml
+++ b/spec/components/schemas/Timelines/CustomerTimelineCustomEvent.yaml
@@ -10,8 +10,14 @@ properties:
   name:
     description: Customer Timeline Custom Event type name. It must not be similar to any Rebilly system event.
     type: string
+    minLength: 1
+    maxLength: 255
   createdTime:
     description: Customer Timeline Custom event created time
+    allOf:
+      - $ref: "#/components/schemas/ServerTimestamp"
+  updatedTime:
+    description: Customer Timeline Custom event updated time
     allOf:
       - $ref: "#/components/schemas/ServerTimestamp"
   _links:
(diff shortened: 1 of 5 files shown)


spec/components/schemas/Timelines/CustomerTimelineCustomEvent.yaml
```diff
@@ -0,0 +1,12 @@
+type: object
+required:
+  - name
+properties:
+  name:
+    description: Customer Timeline Custom event type name. It must not be similar to any Rebilly system event.
+    type: string
+  createdTime:
+    description: Customer Timeline Custom event created time
+    readOnly: true
```
ashkarpetin: Redundant property

spec/components/schemas/Timelines/CustomerTimelineCustomEvent.yaml
```diff
@@ -0,0 +1,11 @@
+type: object
+required:
+  - name
+properties:
+  name:
+    description: Customer Timeline Custom event type name. It must not be similar to any Rebilly system event.
+    type: string
```
adamaltman: `maxLength`? `minLength`?

adamaltman: still missing the min and max here... otherwise, it seems good to me.

spec/components/schemas/Timelines/CustomerTimelineCustomEvent.yaml
```diff
@@ -0,0 +1,29 @@
...
+required:
+  - name
+properties:
+  id:
+    description: Customer Timeline Custom Event identifier string
+    readOnly: true
+    allOf:
+      - $ref: "#/components/schemas/ResourceId"
+  name:
+    description: Customer Timeline Custom Event type name. It must not be similar to any Rebilly system event.
```
slavcodev: What about conflicts with new system events?

2e3s: Probably it should get a prefix or suffix. Then the last remark isn't actual anymore

XaosSintez: As i understood from discussion in Slack, if they created the event type, they can use it further, even if there's new system event with same type.
I'm not sure that prefix/suffix will not confuse our clients. But may be i'm wrong.
@adamaltman just want to make sure you see this discussion :-)

adamaltman: See it?  Yes. 

Understand it?  No.

I understand about conflicts with new system events.  But that's for us to sort out and handle if that happens.

PR #723, opened 2019-07-11
Improve coupons discounts description

VladimirBerdnik:


spec/components/schemas/Coupon/Discounts/percent.yaml
```diff
@@ -9,5 +9,6 @@ allOf:
     properties:
       value:
         description: Discount percent
-        type: number
-        format: double
+        type: integer
+        minimum: 1
```
slavcodev:
```suggestion
        minimum: 0
```

spec/components/schemas/Coupon/Discounts/fixed.yaml
```diff
@@ -12,6 +12,8 @@ allOf:
         description: Discount amount
         type: number
         format: double
+        minimum: 0
+        exclusiveMinimum: true
```
slavcodev:
```suggestion
```

spec/components/schemas/Coupon/Discounts/percent.yaml
VladimirBerdnik: @slavcodev, does it have sence to have such coupons?

spec/components/schemas/Coupon/Discounts/fixed.yaml
adamaltman: Why would you remove this?  It makes a lot of sense.

spec/components/schemas/Coupon/Discounts/percent.yaml
adamaltman: No, I disagree with the suggestion.

spec/components/schemas/Coupon/Discounts/fixed.yaml
slavcodev: I don't see why it cannot be 0. But, I think I understand your point of view.

spec/components/schemas/Coupon/Discounts/percent.yaml
slavcodev: I don't see why we should disallow this case. It does not break any logic of application.

VladimirBerdnik:
So, no notices on this PR? can it be approved?

VladimirBerdnik:
@slavcodev , @adamaltman , do we have agreement on this PR? :)

spec/components/schemas/Coupon/Discounts/percent.yaml
```diff
@@ -9,5 +9,6 @@ allOf:
     properties:
       value:
         description: Discount percent
-        type: number
-        format: double
+        type: integer
```
slavcodev: Is the percentage discount maybe only integer?

spec/components/schemas/Coupon/Discounts/percent.yaml
```diff
@@ -9,5 +9,6 @@ allOf:
...
+        minimum: 1
+        maximum: 100
```
slavcodev: Just curios is the discount 200% posible?

spec/components/schemas/Coupon/Discounts/percent.yaml
VladimirBerdnik: I took this restriction from backend

slavcodev: I think we need to fix backend.

VladimirBerdnik: I actually can't imagine a use case for float percentage for disount

VladimirBerdnik: It wasn't allowed yet and looks like we have no any requests from our clients to have such possibility

2e3s:
I wonder if the maximum percentage is <100

VladimirBerdnik:
> I wonder if the maximum percentage is <100

you mean should be less than 100? I think having 100% discount coupon has sense

2e3s:
Oh, yes. I mean <=100%. 120% discount is a strange thing, credits should be added differently.

2e3s:
Sorry, I missed that it's already there

PR #728, opened 2019-07-25
Add operationId to paths specs

ptrofimov:
I added operationId property to all the paths. Somewhere this attribute already exists. So I did not touch existing ones due to compatibility.

And somewhere I found errors of typos in descriptions, so I am gonna fix them as well.

### DIFF
--- a/spec/paths/3dsecure.yaml
+++ b/spec/paths/3dsecure.yaml
@@ -2,6 +2,7 @@ get:
   tags:
     - 3D Secure
   summary: Retrieve a list of ThreeDSecure entries
+  operationId: get3DSecureEntries
   parameters:
     - $ref: "#/components/parameters/collectionLimit"
     - $ref: "#/components/parameters/collectionOffset"
@@ -33,6 +34,7 @@ post:
   tags:
     - 3D Secure
   summary: Create a ThreeDSecure entry
+  operationId: create3DSecureEntry
   description: |
     Create a ThreeDSecure entry
   requestBody:
--- a/spec/paths/3dsecure.yaml
+++ b/spec/paths/3dsecure.yaml
@@ -2,7 +2,7 @@ get:
   tags:
     - 3D Secure
   summary: Retrieve a list of ThreeDSecure entries
-  operationId: get3DSecureEntries
+  operationId: Get3DSecureCollection
   parameters:
     - $ref: "#/components/parameters/collectionLimit"
     - $ref: "#/components/parameters/collectionOffset"
@@ -34,7 +34,7 @@ post:
   tags:
     - 3D Secure
   summary: Create a ThreeDSecure entry
-  operationId: post3DSecureEntry
+  operationId: Post3DSecure
   description: |
     Create a ThreeDSecure entry
   requestBody:
--- a/spec/paths/3dsecure.yaml
+++ b/spec/paths/3dsecure.yaml
@@ -34,7 +34,7 @@ post:
   tags:
     - 3D Secure
   summary: Create a ThreeDSecure entry
-  operationId: create3DSecureEntry
+  operationId: post3DSecureEntry
   description: |
     Create a ThreeDSecure entry
   requestBody:
--- a/spec/paths/attachments.yaml
+++ b/spec/paths/attachments.yaml
@@ -2,6 +2,7 @@ get:
   tags:
     - Files
   summary: Retrieve a list of Attachments
+  operationId: getAttachments
   description: |
     Retrieve a list of Attachments
   parameters:
@@ -51,6 +52,7 @@ post:
   tags:
     - Files
   summary: Create an Attachment
+  operationId: createAttachment
   description: |
     Create an Attachment
   requestBody:
--- a/spec/paths/attachments.yaml
+++ b/spec/paths/attachments.yaml
@@ -2,7 +2,7 @@ get:
   tags:
     - Files
   summary: Retrieve a list of Attachments
-  operationId: getAttachments
+  operationId: GetAttachmentCollection
   description: |
     Retrieve a list of Attachments
   parameters:
@@ -52,7 +52,7 @@ post:
   tags:
     - Files
   summary: Create an Attachment
-  operationId: postAttachment
+  operationId: PostAttachment
   description: |
     Create an Attachment
   requestBody:
--- a/spec/paths/attachments.yaml
+++ b/spec/paths/attachments.yaml
@@ -52,7 +52,7 @@ post:
   tags:
     - Files
   summary: Create an Attachment
-  operationId: createAttachment
+  operationId: postAttachment
   description: |
     Create an Attachment
   requestBody:
--- a/spec/paths/authentication-options.yaml
+++ b/spec/paths/authentication-options.yaml
@@ -27,7 +27,7 @@ put:
   tags:
     - Customer Authentication
   summary: Change authentication options
-  operationId: updateAuthenticationOptions
+  operationId: putAuthenticationOptions
   description: |
     Change options
   requestBody:
--- a/spec/paths/authentication-options.yaml
+++ b/spec/paths/authentication-options.yaml
@@ -2,6 +2,7 @@ get:
   tags:
     - Customer Authentication
   summary: Read current authentication options
+  operationId: getAuthenticationOptions
   description: |
     Read current authentication options
   responses:
@@ -26,6 +27,7 @@ put:
   tags:
     - Customer Authentication
   summary: Change authentication options
+  operationId: updateAuthenticationOptions
   description: |
     Change options
   requestBody:
--- a/spec/paths/authentication-options.yaml
+++ b/spec/paths/authentication-options.yaml
@@ -2,7 +2,7 @@ get:
   tags:
     - Customer Authentication
   summary: Read current authentication options
-  operationId: GetAuthenticationOptionCollection
+  operationId: GetAuthenticationOption
   description: |
     Read current authentication options
   responses:
--- a/spec/paths/authentication-options.yaml
+++ b/spec/paths/authentication-options.yaml
@@ -27,7 +27,7 @@ put:
   tags:
     - Customer Authentication
   summary: Change authentication options
-  operationId: PutAuthenticationOptionCollection
+  operationId: PutAuthenticationOption
   description: |
     Change options
   requestBody:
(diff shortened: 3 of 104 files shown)


spec/paths/attachments.yaml
```diff
@@ -51,6 +52,7 @@ post:
   tags:
     - Files
   summary: Create an Attachment
+  operationId: createAttachment
```
slavcodev: `create` is not appropriate ter to use, the PUT also may create resource. I believe we should use standard method names `get`, `post`, `put` and so on

spec/paths/payments@{id}.yaml
```diff
@@ -30,6 +31,7 @@ put:
   tags:
     - Payments
   summary: Create a payment with predefined ID
+  operationId: createPaymentWithPredefinedId
```
VladimirBerdnik:
```suggestion
  operationId: createOrUpdatePayment
```
according to style of other items, but maybe better to use 'upsertPayment' (same for similar `createOrUpdate`)? To make operationId shorter?

spec/paths/attachments.yaml
VladimirBerdnik: but in case of `abandonInvoice` and `voidInvoice` having meaningful verb makes sense, I think

slavcodev: `abandonInvoice` and `voidInvoice` looks like RPC style not REST.

VladimirBerdnik: could you please propose operationId for these two methods?
_postInvoiceToAbandoned_ or something like this?

slavcodev: I'm not good in naming, something like `postInvoiceAbandonRequest`

VladimirBerdnik: well, I think the style of naming may become a holy war :)

slavcodev: To avoid another holy war, I propose to use same names like we use for request handlers, eq `PostAttachment`

VladimirBerdnik: agree. According to OpenAPI suggestion: `Some code generators use this value to name the corresponding methods in code`
But what if we will rename our request handlers? We can rename operationId as well, but posted links will be broken

slavcodev: We might will want to rename operations in any cases and links will be broken as well. Links will be broken if we remove any API. I.e. this should not bother about this.

VladimirBerdnik: @ptrofimov , this change probably will be not the most interesting task, but still important. If you agree with this conclusion - please do, if not - we can discuss it with team

ptrofimov: Well, I thought about proper names for IDs, and I tried many options. Why did I choose these?

The most important - ID should be human-readable, and not only by technical specialist. Because these IDs will be displayed for customers in the API log, and every customer should clearly understand them. That's why I declined the variant with HTTP verbs - put, post, patch. The only proper verb is DELETE. And GET of course.

And I realized that PUT can create objects as well. That's why I named put in appropriate cases as createOrUpdate.

spec/paths/payments@{id}.yaml
ptrofimov: yes, I was choosing between upsert (mongo style) and replace (mysql style). But these terms are so technical! How do you think?

spec/paths/attachments.yaml
slavcodev: What do you mean by?
> Because these IDs will be displayed for customers in the API log

spec/paths/payments@{id}.yaml
VladimirBerdnik: I'm not sure that non-technical users will read specification, but I may be wrong

slavcodev: Me either. The API documentation is for API integration implementatin. It's written by thecnical people. Moreother, I don't think the operation names are useful for people who read the doc, the purpose of these names is not for documentation.

spec/paths/attachments.yaml
ptrofimov: This is the task https://jira.ilabsinc.com/browse/GWY-10561 that I am doing it for.

ptrofimov: Yes, in the API log for customers now URLs are used. (like v2.1/customers). They are supposed to be replaced (or extended) by operation IDs (like createCustomer).

slavcodev: I think the `PutSomething` or `PatchSomething` would be fine in API logs as operation ID.

ptrofimov: And we consider HTTP verbs what about PostCustomer or PostBankAccount? It sounds like we are sending customer to somewhere :)

adamaltman: To where?  HTTP verbs seems fine to me.

ptrofimov: OK. I will change to HTTP verbs where it is possible.

ptrofimov:
@VladimirBerdnik @slavcodev @adamaltman I changed to HTTP verbs, please review again.

spec/paths/3dsecure.yaml
```diff
@@ -2,6 +2,7 @@ get:
   tags:
     - 3D Secure
   summary: Retrieve a list of ThreeDSecure entries
+  operationId: get3DSecureEntries
```
slavcodev:
```suggestion
  operationId: get3DSecureEntryCollection
```
Let's use resource name in the singular form, in some cases plural is not avaialable.

ptrofimov:
@slavcodev like this?

spec/paths/authentication-options.yaml
```diff
@@ -2,6 +2,7 @@ get:
   tags:
     - Customer Authentication
   summary: Read current authentication options
+  operationId: GetAuthenticationOptionCollection
```
slavcodev:
```suggestion
  operationId: GetAuthenticationOption
```

spec/paths/authentication-options.yaml
```diff
@@ -26,6 +27,7 @@ put:
   tags:
     - Customer Authentication
   summary: Change authentication options
+  operationId: PutAuthenticationOptionCollection
```
slavcodev:
```suggestion
  operationId: PutAuthenticationOption
```

spec/paths/authentication-options.yaml
slavcodev: This schema is incorrect, the response is not array but it's `AuthenticationOptions`

spec/paths/coupons-redemptions.yaml
```diff
@@ -36,6 +37,7 @@ post:
   tags:
     - Coupons
   summary: Redeem a coupon
+  operationId: PostCouponRedemptionRequest
```
slavcodev:
```suggestion
  operationId: PostCouponRedemption
```
Inconsistent name

spec/paths/coupons-redemptions@{id}@cancel.yaml
```diff
@@ -4,6 +4,7 @@ post:
   tags:
     - Coupons
   summary: Cancel a coupon redemption
+  operationId: PostCouponRedemptionCancellationRequest
```
slavcodev:
```suggestion
  operationId: PostCouponRedemptionCancellationRequest
```
The `Request` suffix is not necesary here.

spec/paths/authentication-tokens@{token}@exchange.yaml
```diff
@@ -9,6 +9,7 @@ post:
   tags:
     - Customer Authentication
   summary: Exchange
+  operationId: PostAuthenticationTokenExchangeRequest
```
slavcodev:
```suggestion
  operationId: PostAuthenticationTokenExchange
```
`Request` is not necesary Request

spec/paths/coupons-redemptions@{id}@cancel.yaml
ptrofimov: Ok ,wo what is the rule, where the suffix is necessary and where it is not?

spec/paths/authentication-options.yaml
ptrofimov: It is array https://github.com/Rebilly/RebillyAPI/blob/2423a0b51cb249f221311d7c0d1f91f6572a2ee0/spec/paths/authentication-options.yaml#L21

spec/paths/coupons-redemptions@{id}@cancel.yaml
slavcodev: I just post my thoughts https://github.com/Rebilly/RebillyAPI/pull/728#pullrequestreview-267660850

spec/paths/authentication-options.yaml
slavcodev: > This schema is incorrect

The backend returns object.

ptrofimov: oh, no. sure, when I was naming methods I supposed that schema is correct.

ptrofimov:
@slavcodev I got rid of Request suffix where it is possible

spec/paths/tags@{tag}@customers.yaml
```diff
@@ -4,6 +4,7 @@ post:
   tags:
   - Tags
   summary: Tag a list of customers
+  operationId: PostTagCustomerCollectionRequest
```
slavcodev:
```suggestion
  operationId: PostTagCustomerCollection
```

adamaltman:
On this:

However we probably can avoid this suffix in all cases, for example
PostInvoiceIssueRequest -> PostInvoiceIssuing
PostInvoiceAbandonRequest -> PostInvoiceAbondanation
PostInvoiceVoidRequest -> PostInvoiceVoiding

There are real noun's we should use:

PostInvoiceIssuance
PostInvoiceAbandonment
PostInvoiceVoid

spec/paths/invoices@{id}@void.yaml
```diff
@@ -4,6 +4,7 @@ post:
   tags:
     - Invoices
   summary: Void an invoice
+  operationId: PostInvoiceVoiding
```
adamaltman: Please note that `void` is a noun.
```suggestion
  operationId: PostInvoiceVoid
```

slavcodev:
@ptrofimov please try to avoid `force pushes` to PRs in which participate other teammates, especially after reviews. It's annoying to get back and re-review all again.

ptrofimov:
@slavcodev I force pushed only the last commit that you have not seen before, because it was something strang with author info of commit

PR #730, opened 2019-08-16
Add invoice reissue endpoint

XaosSintez:


spec/paths/invoices@{id}@reissue.yaml
```diff
@@ -0,0 +1,29 @@
+parameters:
+  - $ref: "#/components/parameters/resourceId"
+post:
+  tags:
+    - Invoices
+  summary: Reissue an invoice
+  operationId: PostInvoiceReissuance
+  description: |
+    Reissue an invoice with specified identifier string. It must be issued (status must be `unpaid` or `past-due`).
```
2e3s: When such a reissued invoice becomes past due?

XaosSintez: I will add `dueTime` to request body

2e3s: We don't read `dueTime` for issuing invoice by the way

PR #731, opened 2019-08-16
Add revision to Subscription resource

ashkarpetin:


spec/components/schemas/Subscription.yaml
```diff
@@ -161,6 +161,12 @@ properties:
     nullable: true
     example: PO123456
     type: string
+  revision:
+    description: |
+      The number of times the transaction data has been modified.
```
XaosSintez:
```suggestion
      The number of times the subscription data has been modified.
```

dara123: `transaction` or `subscription` ?

PR #746, opened 2019-10-02
Add Coupon paid-by-time restriction type

XaosSintez:


spec/components/schemas/Coupon/CouponRestriction.yaml
```diff
@@ -15,5 +15,6 @@ properties:
       - restrict-to-subscriptions
       - restrict-to-products
       - total-redemptions
+      - pay-by-when
```
2e3s: Maybe there can be a clearer name, all other restrictions are easily understood. What does it do?

XaosSintez: It means that if invoice will be paid before `byWhen` time, it will get some configured discount.
I just took restriction type name proposed by Adam. If you will propose better name,  i will appreciate it :-)

2e3s: `paid-by-time` would be clearer, probably.

2e3s: Wouldn't it have to be something more relative? Like 10 days before renewal?

XaosSintez: > `paid-by-time`

Looks ok to me. Will use this name

> Wouldn't it have to be something more relative? Like 10 days before renewal or due time? I wonder if you'd have to create a redemption for every renewal manually

I think every redemption will have its own restriction for each case. Please take a look at the Jira ticket, there is some related discussion

spec/components/schemas/Coupon/CouponRestrictions/paid-by-time.yaml
```diff
@@ -0,0 +1,11 @@
...
+allOf:
+  - $ref: "#/components/schemas/CouponRestriction"
+  - $ref: "#/components/schemas/RedemptionRestriction"
+  - type: "object"
+    required:
+     - "time"
+    properties:
+      time:
+        type: string
+        format: date-time
```
2e3s:
```suggestion
        format: date-time
        description: The time when the coupon's redemption is no longer valid and removed from the invoice.
```

PR #747, opened 2019-10-02
Add invoice recalculation endpoint

XaosSintez:


spec/paths/invoices@{id}@recalculate.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - $ref: "#/components/parameters/resourceId"
+post:
+  tags:
+    - Invoices
+  summary: Recalculate an invoice
+  operationId: PostInvoiceRecalculation
+  description: |
+    Recalculate an invoice with specified identifier string.
+  responses:
+    201:
```
ashkarpetin: Should be 200?

XaosSintez: No, i think it should be 201. See similar endpoints like `invoices/{id}/void`, `invoices/{id}/issue`

2e3s:
Missing the link in invoice resource

XaosSintez:
@2e3s Which link?

adamaltman:
I think he means you can add a link from the invoice itself to the recalculate operation.

adamaltman:
(I clicked close by accident.)

spec/paths/invoices@{id}@recalculate.yaml
2e3s: > The HTTP 201 Created success status response code indicates that the request has succeeded and has led to the creation of a resource. The new resource is effectively created before this response is sent back and the new resource is returned in the body of the message, its location being either the URL of the request, or the content of the Location header.

spec/paths/invoices@{id}@recalculate.yaml
```diff
@@ -0,0 +1,29 @@
+parameters:
+  - $ref: "#/components/parameters/resourceId"
+post:
+  tags:
+    - Invoices
+  summary: Recalculate an invoice
+  operationId: PostInvoiceRecalculation
+  description: |
+    Recalculate an invoice with specified identifier string.
```
adamaltman: This description is pretty anemic.

You may want to say that this includes, discounts, shipping and taxes.

You may also want to provide some examples of when you would want to use it.  For example, if you revoke a coupon, you may want to recalculate.

PR #750, opened 2019-10-04
Add invoice due time shift settings to Subscription and Plan entities

ptrofimov:


spec/components/schemas/Invoices/InvoiceDueTimeShift.yaml
```diff
@@ -0,0 +1,18 @@
...
+description: |
+  The `due time shift` in conjunction with `billing anchor` settings allows to setup different billing use cases such as:
+
+  - Issue invoice immediately when the service period _starts_ and bill it immediately
+  - Issue invoice immediately when the service period _starts_ and bill it after _interval of time_
+  - Issue invoice _interval of time_ before the service period _starts_ and bill it after _interval of time_
+  - Issue invoice _interval of time_ after the service period _starts_ and bill it after _interval of time_
+properties:
+  # This should be specified in each schedule instruction method, for example "immediately" don't need it.
+  # But due some limitations of the Swagger v2 we must added it here.
```
adamaltman: We use OAS3 though, not v2.

spec/components/schemas/Invoices/InvoiceDueTimeShift.yaml
```diff
@@ -0,0 +1,18 @@
...
+  chronology:
+    type: string
+    description: The chronology of the invoice due time relatively to the invoice issue time.
```
adamaltman:
```suggestion
    description: The chronology of the invoice due time relative to the invoice issue time.
```

spec/components/schemas/Scheduling/InvoiceDueTimeScheduleInstruction.yaml
```diff
@@ -0,0 +1,19 @@
+type: object
+description: |
+  The calculation instruction of invoice due time shift.
+  This is used in conjunction with the **billing anchor settings** to calculate the time the invoice gets paid.
```
adamaltman:
```suggestion
  This is used in conjunction with the **billing anchor settings** to calculate the invoice due time.
```
It's not necessarily paid unless autopay is enabled.

ptrofimov: right! thx

spec/components/schemas/Subscription.yaml
```diff
@@ -92,6 +92,15 @@ properties:
     example: null
     allOf:
       - $ref: "#/components/schemas/PlanBillingAnchor"
+  invoiceDueTimeShift:
+    description: |
+      The time when invoice gets paid is by default +1 hour from issue time.
+      With this setting you can shift this time.
+      Due time shift is relative to issue time that could be shifted by billing anchor settings.
```
adamaltman:
```suggestion
      Due time shift is relative to issued time that could be shifted by billing anchor settings.
```

spec/components/schemas/Invoices/InvoiceDueTimeShift.yaml
ptrofimov: I can remove this comment

spec/components/schemas/Subscription.yaml
```diff
@@ -92,6 +92,15 @@ properties:
     example: null
     allOf:
       - $ref: "#/components/schemas/PlanBillingAnchor"
+  invoiceDueTimeShift:
+    description: |
+      The time when invoice gets paid is by default +1 hour from issue time.
```
2e3s: Maybe add a default value? I saw that OpenAPI supports complex types

ptrofimov: Default value is here https://github.com/Rebilly/RebillyAPI/pull/750/files#diff-3799c3d3df599168b866e950132b6130R8

2e3s:
Maybe this should live in Plan in the first place and be overridden in Subscription (if provided)?

spec/components/schemas/Invoices/InvoiceDueTimeShift.yaml
```diff
@@ -0,0 +1,16 @@
...
+  - Issue invoice immediately when the service period _starts_ and bill it after _interval of time_
+  - Issue invoice _interval of time_ before the service period _starts_ and bill it after _interval of time_
+  - Issue invoice _interval of time_ after the service period _starts_ and bill it after _interval of time_
+properties:
+  chronology:
+    type: string
+    description: The chronology of the invoice due time relative to the invoice issue time.
+    enum:
+      - "after"
+  invoiceDueTimeShiftInstruction:
```
creocoder: Why you use `invoiceDueTimeShift` prefix for property there? Looks very inconsistent. Please fix that.

spec/components/schemas/Invoices/InvoiceDueTimeShift.yaml
```diff
@@ -0,0 +1,16 @@
+type: object
+description: |
+  The `due time shift` in conjunction with `billing anchor` settings allows to setup different billing use cases such as:
+
+  - Issue invoice immediately when the service period _starts_ and bill it immediately
+  - Issue invoice immediately when the service period _starts_ and bill it after _interval of time_
+  - Issue invoice _interval of time_ before the service period _starts_ and bill it after _interval of time_
+  - Issue invoice _interval of time_ after the service period _starts_ and bill it after _interval of time_
+properties:
+  chronology:
```
creocoder: If only chronology `after` make sense why ever you need that in payload???

creocoder:
Optimized payload which make sense:

```
{
    ...,
    "invoiceDueTimeShift": {
        "method": ...,
        "duration": ...,
        "unit": ...
    },
    ...
}
```

spec/components/schemas/Scheduling/InvoiceDueTimeScheduleInstruction.yaml
```diff
@@ -0,0 +1,19 @@
...
+  "duration": 1
+  "unit": "hour"
+required:
+  - method
+properties:
+  method:
+    type: string
+    default: "date-interval"
+    enum:
+      - "immediately"
```
creocoder: This does not make sense.

creocoder:
Seems `method` also is `date-inverval` and does not make sense in that case in payload. Optimized payload now looks like:

```
{
    ...,
    "invoiceDueTimeShift": {
        "duration": ...,
        "unit": ...
    },
    ...
}
```

2e3s:
@creocoder defining `null` as meaning of immediately?

creocoder:
@2e3s What do you mean? No `invoice-due-time-shift` in payload === no invoice due time shift.

ptrofimov:
ok, i see so many questions here. I can suggest another version - to combine billing anchor settings and invoice due time shift settings. may be it will be simpler and better.

2e3s:
> No `invoiceDueTimeShift` in payload === no invoice due time shift.

I'd like to notice that the default should be `dueTime = issuedTime + 1 hour`.

spec/components/schemas/Plans/PlanBillingAnchor.yaml
```diff
@@ -19,3 +17,5 @@ properties:
       - "after"
   billingAnchorInstruction:
```
adamaltman: I think this would be even more clear if we renamed this to `issueTimeShiftInstruction` (and I believe we can do that as the only client to adjust is Recomm).

spec/components/schemas/Scheduling/DueTimeShiftInstruction.yaml
```diff
@@ -0,0 +1,18 @@
+type: object
+description: |
+  The calculation instruction of due time.
+  This is used in conjunction with the **billing anchor** to calculate due time of invoice.
+  The chronology of due time shift is always **after**.
```
adamaltman: If we can rename the billingAnchorInstruction, then this may make more sense like this:
```suggestion
  The chronology of due time shift is always **after** the issuance of the invoice.
```

spec/components/schemas/Scheduling/DueTimeShiftInstruction.yaml
```diff
@@ -0,0 +1,18 @@
+type: object
+description: |
+  The calculation instruction of due time.
+  This is used in conjunction with the **billing anchor** to calculate due time of invoice.
```
adamaltman:
```suggestion
  This is used in conjunction with the issue time (calculated from the **billing anchor**) to calculate due time of invoice.
```

spec/components/schemas/Plans/PlanBillingAnchor.yaml
ptrofimov: Only one client to adjust? But how can we know that? Api is public.

2e3s: 99% possibility that the settings are not used by anything but our UI. But it can be known from our internal reports.

ptrofimov: This property will be renamed in special task that I created for it, becuase it is not related to the current task and includes recomm and SDK changes.

PR #757, opened 2019-10-21
Fix payment card requestBody on patch

ashkarpetin:


spec/paths/payment-cards@{id}.yaml
```diff
@@ -43,10 +43,10 @@ patch:
               type: string
             expMonth:
               description: Card's expiry month
-              type: string
+              type: integer
```
colinbird: What about "02" ?

ashkarpetin: I'll improve validator

PR #765, opened 2019-10-29
Add support for marking KYC documents as reviewed

Teknologica:
This PR updates only the KYC Document definitions.

- Adds a new `reviewed` boolean that can be toggle through `patch` or included in the initial `put`. 
- Also adds `reviewedTime` to determine when the last review change occurred.
- Moves the KYC Document files into `/KycDocument`
- Adds a `reviewerId` write-only value when updating or creating the document
- Adds a `reviewer` read-only object with `{id, name}` keys that returns the full name filled up from the last provided `reviewerId`


spec/components/schemas/KycDocument/KycDocument.yaml
```diff
@@ -32,6 +32,10 @@ properties:
       - in-progress
       - accepted
       - rejected
+  reviewed:
```
adamaltman: It seems we're missing who reviewed it, which seems really important.

spec/paths/kyc-documents@{id}.yaml
```diff
@@ -31,6 +31,40 @@ get:
       $ref: "#/components/responses/AccessForbidden"
     404:
       $ref: "#/components/responses/NotFound"
+patch:
+  tags:
+    - KYC Documents
+  summary: Update a KYC document with predefined ID
```
adamaltman:
```suggestion
  summary: Update a KYC document
```

spec/paths/kyc-documents@{id}.yaml
```diff
@@ -31,6 +31,40 @@ get:
...
+  operationId: PatchKycDocument
+  description: |
+    Update a KYC document with predefined identifier string
```
adamaltman: What is the reason this would be useful? From what i can see, it is only used to mark something as reviewed (true or false).  I think it would be more useful if the description explains what it can be used for.

Teknologica: Updated the description

spec/components/schemas/KycDocument/KycDocument.yaml
Teknologica: I looked into that a bit and came up with a version similar to what we have to segments.

Teknologica:
devs please re-review

mclassic:
Discussed with @Teknologica and looks fine. He explained to me the reason for `reviewed` as a boolean in the response instead of using `reviewedTime` as a boolean is because `reviewedTime` will just basically be a "last reviewed or unreviewed" time. So we'll still need a boolean to represent the toggle.

Teknologica:
devs please re-review

spec/components/schemas/KycDocument/KycDocument.yaml
```diff
@@ -32,6 +32,29 @@ properties:
...
+    description: Defines whether the document was manually inspected and reviewed
+    type: boolean
+    default: false
+  reviewerId:
+    description: The current reviewer's user ID
+    type: string
+    writeOnly: true
```
adamaltman: I think we would want this to be `readOnly`.  It will be written based on who is logged in and clicks on the "reviewed" button, for example.  So, it's possible that we don't need it here at all, since you have "reviewer" just underneath and that contains the user's ID too.

spec/components/schemas/KycDocument/PatchKycDocument.yaml
```diff
@@ -0,0 +1,16 @@
...
+required:
+  - reviewed
+  - reviewerId
+properties:
+  reviewed:
+    description: |
+      Defines whether the document was manually inspected and reviewed after the automated KYC process. Can be toggled multiple times as it goes through review.
+    type: boolean
+    default: false
+  reviewerId:
```
adamaltman: I think we can skip this, since it would be better detected by the authentication headers.

spec/components/schemas/KycDocument/PatchKycDocument.yaml
```diff
@@ -0,0 +1,16 @@
+type: object
+required:
+  - reviewed
+  - reviewerId
+properties:
```
adamaltman: How about the status: eg `accept` and `reject`?

spec/paths/kyc-documents@{id}.yaml
```diff
@@ -31,6 +31,40 @@ get:
...
+    - KYC Documents
+  summary: Review a KYC document
+  operationId: PatchKycDocument
+  description: |
+    Allows you to toggle the reviewed state of the KYC document when doing manual verification and store the reviewer's information.
+  requestBody:
+    content:
+      application/json:
+        schema:
+          $ref: "#/components/schemas/PatchKycDocument"
```
adamaltman: I find it odd that I can only patch the `reviewed` property when other properties are mutable at this path.  Maybe the path should change to `kyc-documents@{id}@reviews` ?

adamaltman:
I requested reviews from @vykhrystiuk (he is refactoring KYC at the moment on the backend) and @slavcodev (he brings unique API design opinions).

mclassic:
Based on the requirements @Teknologica mentioned to me it seemed ok at the time, but it sounds like that's not the actual case, (i.e. i was told one could specify the reviewer, that it wasn't necessarily the logged in reviewer, etc.)

Teknologica:
> 
> 
> Based on the requirements @Teknologica mentioned to me it seemed ok at the time, but it sounds like that's not the actual case, (i.e. i was told one could specify the reviewer, that it wasn't necessarily the logged in reviewer, etc.)

Yeah just miscommunication when I explained the requirement from my part. What Adam is saying is correct in this case.

spec/components/schemas/KycDocument/PatchKycDocument.yaml
vykhrystiuk: I think that makes sense, cause the reviewer can reject it and it would probably be better to determine that (not only `false`, because it's a default), also we can provide statuses like `pending` for review etc. if we need it

spec/components/schemas/KycDocument/KycDocument.yaml
vykhrystiuk: Do we need a `reviewerId ` field? Are we going to make it possible not only by current user (which we can define by `getAuthenticatedUser`)?

Teknologica: I'm removing the `reviewerId` since it can be implicitly provided when making the request.

spec/components/schemas/KycDocument/PatchKycDocument.yaml
Teknologica: There's already a `/accept` and `/reject` action only for the status. I'm adding `/review` to match the pattern here. I think the status toggle could remain as part of the original actions.

spec/components/schemas/KycDocument/PatchKycDocument.yaml
Teknologica: I learned something new

spec/paths/kyc-documents@{id}.yaml
Teknologica: Added `kyc-documents@{id}@review` to match `kyc-documents@{id}@acceptance` and `kyc-documents@{id}@rejection` .

Teknologica:
devs please re-review

spec/paths/kyc-documents@{id}@review.yaml
```diff
@@ -0,0 +1,36 @@
+parameters:
+  - $ref: "#/components/parameters/resourceId"
+patch:
+  tags:
+    - KYC Documents
+  summary: Review a KYC document
+  operationId: PatchKycDocument
+  description: |
+    Allows you to toggle the reviewed state of the KYC document when doing manual verification and store the reviewer's information.
```
adamaltman: I'm curious though.  If someone uses the accept/reject button manually, wouldn't we consider that is reviewed too?  It seems like it's a more explicit change of the status.  Whereas, clicking "review" is more of an acknowledgement that it was reviewed and there is no change to make to the status.  It seems we would want both to be considered as "reviewed"?

spec/components/schemas/KycDocument/KycDocument.yaml
```diff
@@ -32,6 +32,23 @@ properties:
       - in-progress
       - accepted
       - rejected
+  reviewed:
+    description: Defines whether the document was manually inspected and reviewed
+    type: boolean
+    default: false
```
vykhrystiuk: I'm thinking about how can we tell the difference between `rejected` and `not reviewed` document. `not reviewed` - (`reviewed` = `false`, `reviewedTime` = `null`), `rejected` - (`reviewed` = `false`, `reviewedTime` = `some dateTime`), if I get this rigth.
I wonder if we add property `reviewStatus` instead of `reviewed` with statuses (not reviewed|approved|rejected), for example. Does it make sense?

spec/paths/kyc-documents@{id}@review.yaml
```diff
@@ -0,0 +1,36 @@
+parameters:
+  - $ref: "#/components/parameters/resourceId"
+patch:
+  tags:
+    - KYC Documents
+  summary: Review a KYC document
+  operationId: PatchKycDocument
```
slavcodev:
```suggestion
  operationId: PostKycDocumentReview
```

Please leave `PatchKycDocument` reserved for PATCH to parent endpoint.

spec/components/schemas/KycDocument/KycDocument.yaml
slavcodev: No, it doesn't for me IMO. It will be destinguish by `reviewedTime`. What is the value of new proeprty `reviewStatus`?

slavcodev: I don’t even think that we need “revised”, because it repeats “revised time! = False”, I would not inflate the properties of the object unnecessarily.

spec/paths/kyc-documents@{id}@review.yaml
```diff
@@ -0,0 +1,36 @@
+parameters:
+  - $ref: "#/components/parameters/resourceId"
+patch:
```
slavcodev:
```suggestion
post:
```

spec/components/schemas/KycDocument/KycDocumentReview.yaml
```diff
@@ -0,0 +1,9 @@
+type: object
+required:
+  - reviewed
+properties:
+  reviewed:
+    description: |
+      Defines whether the document was manually inspected and reviewed after the automated KYC process. Can be toggled multiple times as it goes through review.
+    type: boolean
+    default: false
```
slavcodev: What does `reviewd = false` means? Can KYC doc be `reviewed` and later `un-reviewed`?

spec/components/schemas/KycDocument/KycDocument.yaml
vykhrystiuk: I think for clear understanding `not reviewed`, `approved` or `rejected`. Because `rejected` and `approved` mean as reviewed and we should mark `reviewed` as `true` in those cases.

adamaltman: Review relates to the status.  For example, automated analysis may set the status as `rejected`, but a human may review it and decide it can be `accepted`.

However, upon looking at this and thinking again, I wonder what is the use case for changing something from `reviewed: true` to `reviewed: false`?  Why not just have a `reviewedTime` if we cannot change it back to `false`?  

How can we filter in the api by `____Time` is null? (If I recall, we don't support that as a filter right now -- and it would be important if there is no `reviewed` boolean.)

spec/components/schemas/KycDocument/KycDocumentReview.yaml
adamaltman: I just posted this similar question here: https://github.com/Rebilly/RebillyAPI/pull/765#discussion_r343086205

I can't envision this use case.

slavcodev: Yуs. Me either that pushed me and I proposed changes to this API in JIRA

spec/components/schemas/KycDocument/KycDocument.yaml
mclassic: FYI, as of last night we do support `___Time` is null now, and is supported in Recomm now. i.e. `filter=reviewedTime:null` would return all with a null `reviewedTime`

Teknologica:
Updated to match discussion that took place.

Teknologica:
devs please re-review

spec/components/schemas/KycDocument/KycDocument.yaml
```diff
@@ -48,6 +61,11 @@ properties:
     description: Processing date/time
     allOf:
       - $ref: "#/components/schemas/ServerTimestamp"
+  reviewedTime:
```
slavcodev:
```suggestion
  reviewTime:
```

spec/paths/kyc-documents@{id}@review.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - $ref: "#/components/parameters/resourceId"
+post:
+  tags:
+    - KYC Documents
+  summary: Review a KYC document
+  operationId: PostKycDocumentReview
+  description: |
+    Mark the KYC document as reviewed without changing its status. Updates the reviewed time and reviewer information.
+  responses:
+    200:
```
slavcodev:
```suggestion
    201:
```

spec/components/schemas/KycDocument/KycDocument.yaml
```diff
@@ -48,6 +61,11 @@ properties:
     description: Processing date/time
     allOf:
       - $ref: "#/components/schemas/ServerTimestamp"
+  reviewTime:
+    description: Date and time of last manual review
+    readOnly: true
```
adamaltman: The server timestamp is already `readOnly` (I think).

spec/paths/kyc-documents@{id}@acceptance.yaml
```diff
@@ -6,7 +6,7 @@ post:
   summary: Accept a KYC document
   operationId: PostKycDocumentAcceptance
   description: |
-    Manually override automated status and accept the document
+    Manually override automated status and accept the document. Updates the reviewed time and reviewer information.
```
adamaltman:
```suggestion
    Marks that status of the document as `accepted`. Updates the review time and reviewer information.  Intended to be used for manual overrides.
```

spec/paths/kyc-documents@{id}@rejection.yaml
```diff
@@ -6,7 +6,7 @@ post:
   summary: Reject a KYC document
   operationId: PostKycDocumentRejection
   description: |
-    Manually override automated status and reject the document
+    Manually override automated status and reject the document. Updates the reviewed time and reviewer information.
```
adamaltman:
```suggestion
    Marks that status of the document as `accepted`. Updates the review time and reviewer information.  Intended to be used for manual overrides.
```

spec/paths/kyc-documents@{id}@review.yaml
```diff
@@ -0,0 +1,29 @@
+parameters:
+  - $ref: "#/components/parameters/resourceId"
+post:
+  tags:
+    - KYC Documents
+  summary: Review a KYC document
+  operationId: PostKycDocumentReview
+  description: |
+    Mark the KYC document as reviewed without changing its status. Updates the reviewed time and reviewer information.
```
adamaltman:
```suggestion
    Mark the KYC document as reviewed without changing its status. Updates the review time and reviewer information.
```

mclassic:
devs please review

spec/paths/kyc-documents@{id}@review.yaml
```diff
@@ -0,0 +1,36 @@
+parameters:
+  - $ref: "#/components/parameters/resourceId"
+post:
+  tags:
+    - KYC Documents
+  summary: Review a KYC document
+  operationId: PostKycDocumentReview
+  description: |
+    Mark the KYC document as reviewed without changing its status. Updates the review time and reviewer information.
```
ashkarpetin: Can you explain why we need to review document without changing it status?

mclassic: BAd wording I'll fix

spec/paths/kyc-documents@{id}@review.yaml
```diff
@@ -0,0 +1,36 @@
+parameters:
+  - $ref: "#/components/parameters/resourceId"
+post:
+  tags:
+    - KYC Documents
+  summary: Review a KYC document
+  operationId: PostKycDocumentReview
+  description: |
+    Mark the KYC document as reviewed. Updates the review time and reviewer information.
```
ashkarpetin: Still not clear to me, what is reviewed KYC document? Is status changed to accepted?

spec/paths/kyc-documents@{id}@rejection.yaml
```diff
@@ -6,7 +6,7 @@ post:
   summary: Reject a KYC document
   operationId: PostKycDocumentRejection
   description: |
-    Manually override automated status and reject the document
+    Marks that status of the document as `accepted`. Updates the review time and reviewer information.  Intended to be used for manual overrides.
```
edtoken: probably `accepted` -> `rejected`

mclassic:
```suggestion
    Marks that status of the document as `rejected`. Updates the review time and reviewer information.  Intended to be used for manual overrides.
```

spec/paths/kyc-documents@{id}@review.yaml
mclassic: If they change it to accepted sure. The choices are displayed in the request schema.

spec/components/schemas/KycDocument/KycDocumentReviewer.yaml
```diff
@@ -0,0 +1,24 @@
+type: object
+description: Details of the last reviewer including his name and id
```
vykhrystiuk: I just noticed, missed period at the end of the descriptions.

spec/components/schemas/KycDocument/KycDocumentReviewer.yaml
```diff
@@ -0,0 +1,24 @@
...
+    description: Reviewer's user ID
+    type: string
+    readOnly: true
+    allOf:
+      - $ref: '#/components/schemas/ResourceId'
+  name:
+    description: Reviewer's first and last name
+    type: string
+    readOnly: true
+  reviewedTime:
```
slavcodev: I think it would be better to rename in to `reviewTime`

spec/components/schemas/KycDocument/KycDocumentReviewer.yaml
```diff
@@ -0,0 +1,24 @@
+type: object
+description: Details of the last reviewer including his name and id
+readOnly: true
+properties:
+  id:
+    description: Reviewer's user ID
+    type: string
+    readOnly: true
```
slavcodev: I guess no need if root object is fully read-only

slavcodev: Actualy I see this is used as request body, so I'm confused how the API is supposed to work.

mclassic: I'll remove the root object `readOnly` in an upcoming PR. It doesn't affect the properties inside it.

spec/components/schemas/KycDocument/KycDocumentReviewer.yaml
mclassic: I renamed it to `reviewedTime` purposely to match tense of all other timestamps in our system.

mclassic: i'll deal with this in new PR now

spec/components/schemas/KycDocument/KycDocument.yaml
```diff
@@ -32,6 +32,12 @@ properties:
       - in-progress
       - accepted
       - rejected
+  reviewers:
+    type: array
```
adamaltman: I don't think this is a good idea.  It also conflicts with the description below.

spec/components/schemas/KycDocument/KycDocumentReviewer.yaml
adamaltman: @slavcodev is correct IMO.

spec/components/schemas/KycDocument/KycDocument.yaml
adamaltman: It was discussed somewhere, but I can't find where.  I think it adds unnecessary complexity to the design without benefit.

PR #766, opened 2019-10-29
Move subscription link from invoice item to invoice

2e3s:
All Invoice Items can belong only to 1 subscription as the invoice.


spec/components/schemas/Links/SubscriptionLink.yaml
```diff
@@ -6,6 +6,6 @@ properties:
     description: The link type
     type: string
     enum:
-      - subscription
+      - order
```
2e3s:
```suggestion
      - subscription
```

PR #772, opened 2019-11-14
Add too many requests

king555123:
We should be able to return the retry-after header in all responses.

### DIFF
--- a/spec/components/responses/TooManyRequests.yaml
+++ b/spec/components/responses/TooManyRequests.yaml
@@ -1,0 +1,6 @@
+description: 'You sent too many requests in a given amount of time'
+headers:
+  Retry-After:
+    description: The number of seconds to wait until next try
+    schema:
+      type: integer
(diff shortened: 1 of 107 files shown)


spec/components/responses/TooManyRequests.yaml
```diff
@@ -0,0 +1,6 @@
+description: 'You sent too many requests in a given amount of time'
+headers:
+  Retry-After:
+    description: The number of seconds to wait until next try
+    schema:
+      type: integer
```
XaosSintez: Are you sure this is a correct schema? Shouldn't it be `object`?

slavcodev: @XaosSintez, nope. This is schema for header value. However, by HTTP spec this header can have the `Date` too, not only integer.

mvphero:
Will be re-do from scratch

PR #776, opened 2019-11-27
Make bank account status read-only

ashkarpetin:


adamaltman:
How does someone deactivate a bank account?

adamaltman:
@dara123 I don't understand how you can give this a checkmark given your experience with a particular use case.

ashkarpetin:
@adamaltman You can create bank account and deactivate it

adamaltman:
This won't work with a specific client integration that is already existing and in-use.

PR #781, opened 2019-12-13
Add a KYC document and Customer collection endpoint

mclassic:
Add an endpoint that will return a collection of KYC Documents and the Customer to which they belong.


spec/components/schemas/KycDocumentsAndCustomer.yaml
```diff
@@ -0,0 +1,21 @@
...
+      allOf:
+        - $ref: "#/components/schemas/KycDocument"
+  _links:
+    type: array
+    description: The links related to resource
+    readOnly: true
+    minItems: 1
+    items:
+      anyOf:
+      - $ref: "#/components/schemas/SelfLink"
```
ashkarpetin: What does the link point to?

mclassic: I don't know. I thought this was what was in all our API responses. I am more than willing to remove it, but I think it points to itself?

mclassic: Oh I see what you mean, it's in the wrong area. Sorry, my fault.

PR #786, opened 2019-12-17
Make bank account status read only

ashkarpetin:
In PR  #783 I forgot to remove status from request body


2e3s:
But it removes the fields instead of making them read only?

ashkarpetin:
@2e3s It's a request schema (see preview)

PR #789, opened 2019-12-19
Introduce Storefront Offers

justnero:


spec/components/requestBodies/Offer.yaml
```diff
@@ -0,0 +1,32 @@
...
+      type: object
+      required:
+        - currency
+        - websiteId
+        - charges
+      properties:
+        currency:
+          allOf:
+            - $ref: "#/components/schemas/CurrencyCode"
+        websiteId:
```
adamaltman: Why is this required and not just a possible restriction?

spec/components/schemas/Offers/Restrictions/total-purchases.yaml
```diff
@@ -0,0 +1,11 @@
+allOf:
+  - $ref: "#/components/schemas/OfferRestriction"
+  - type: object
+    required:
+      - amount
+    properties:
+      amount:
+        description: The number of purchases allowed on the offer
```
adamaltman: This whole restriction part doesn't make sense to me.

spec/components/schemas/Offers/OfferRestriction.yaml
```diff
@@ -0,0 +1,11 @@
...
+discriminator:
+  propertyName: type
+required:
+  - type
+properties:
+  type:
+    type: string
+    enum:
+      - total-purchases
+      - gateway-accounts
```
adamaltman: This doesn't make sense.  I need to re-tell the use case.

spec/paths/offers.yaml
```diff
@@ -0,0 +1,54 @@
...
+              $ref: "#/components/schemas/Offer"
+    401:
+      $ref: "#/components/responses/AccessForbidden"
+post:
+  tags:
+    - Offers
+  summary: Create an Offer
+  operationId: PostOffer
+  description: |
+    Create an Offer
```
adamaltman: This is where you would write in human-readable sentences and explain why someone would want to create an offer. And you should add the tag to the tags section in openapi.yaml with a description including the what/why explained there.

spec/components/requestBodies/Offer.yaml
```diff
@@ -0,0 +1,32 @@
...
+          description: The website's ID
+          allOf:
+            - $ref: "#/components/schemas/ResourceId"
+        charges:
+          description: The charges applied customer
```
adamaltman: This seems an incorrect description.  As I understand the customer may select one of these.

spec/components/schemas/Offers/OfferCharge.yaml
```diff
@@ -0,0 +1,33 @@
+type: object
+required:
+  - name
+  - pricing
+  - quantity
+properties:
+  name:
+    type: string
+    description: The name of the charge (used in integrations only)
```
adamaltman: Does it need to be unique? How is it used in integrations?

spec/components/schemas/Offers/OfferRestriction.yaml
```diff
@@ -0,0 +1,11 @@
+type: object
+discriminator:
```
adamaltman: When possible, avoid the discriminator. I will request every API that has it should use `oneOf` instead.

spec/paths/offers.yaml
adamaltman: I forgot to mention, and when you want to do it.

spec/components/requestBodies/Offer.yaml
justnero: No, all charges are applied at the same time. The only thing customer can choose is quantity

adamaltman: I don't understand how it works.  That means this documentation is incomplete.

spec/components/requestBodies/Offer.yaml
justnero: It doesn't make sense for `websiteId` to be optional here. When it is required we can use it for transaction creation, but as an optional value it doesn't worth. For limiting Offer visibility on different websites we have ACL

adamaltman: I think you misunderstand me here.  I'm not talking about making this optional.  I'm talking about removing it completely -- and then adding a possible restriction type for website.  As you propose, this solution will not work for certain clients.  I'm happy to discuss it on slack.

spec/paths/offers@{id}.yaml
```diff
@@ -0,0 +1,92 @@
...
+      $ref: "#/components/responses/NotFound"
+    422:
+      $ref: "#/components/responses/InvalidDataError"
+delete:
+  tags:
+    - Offers
+  summary: Delete an Offer
+  operationId: DeleteOffer
+  description: |
+    Delete an Offer with predefined identifier string
```
ashkarpetin: Does delete remove the record from table completely or it just changes status to deactivated?

justnero:
Offers are not relevant for our merchants in current version

PR #790, opened 2020-01-10
Add Organization-Id header

iamdevice:
Add Organization-Id header to query header parameters


spec/components/parameters/organizationId.yaml
```diff
@@ -0,0 +1,6 @@
+name: Organization-Id
+in: header
+description: Organization identifier (if not specified, the default organization will be used)
```
VladimirBerdnik:
```suggestion
description: Organization identifier in scope of which need to perform request (if not specified, the default organization will be used)
```

Not sure about the sentence itself, but the idea is to describe the parameter more verbose and not to have description is just the same as parameter name

iamdevice:
> I think this PR would be a pain to re-do manually after we upgrade the structure. I propose we merge it before we upgrade.

Sounds good

PR #795, opened 2020-01-15
Add write-only `Invoice.couponIds`

edmandiesamonte:
Support for redeeming coupons to be used to a specific invoice.

### DIFF
--- a/spec/components/schemas/Invoices/Invoice.yaml
+++ b/spec/components/schemas/Invoices/Invoice.yaml
@@ -93,6 +93,25 @@ properties:
     items:
       allOf:
         - $ref: "#/components/schemas/InvoiceTax"
+  couponIds:
+    type: array
+    nullable: true
+    description: |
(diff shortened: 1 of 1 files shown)


spec/components/schemas/Invoices/Invoice.yaml
```diff
@@ -93,6 +93,25 @@ properties:
...
+    description: |
+      A list of coupons to redeem on the customer and restrict to this invoice.
+      Read more about [coupons here](https://help.rebilly.com/invoices-and-subscriptions/coupons-discounts).
+
+      This parameter respects the following logic:
+
+      - When not passed then applied coupons will not be changed.
+
+      - When empty array passed then all applied coupon redemptions will be canceled.
+
```
2e3s: In markdown there is no need in those empty lines before list points

edmandiesamonte: Thanks, @2e3s ! :)

PR #796, opened 2020-01-16
Update to new structure

adamaltman:


openapi/openapi.yaml
```diff
@@ -0,0 +1,489 @@
...
+      - Taxes
+  - name: Misc
+    tags:
+      - Reports
+      - Search
+      - Users
+security:
+  - SecretApiKey: []
+  - JWT: []
+paths:
```
adamaltman: This is one of the key changes, team.  There are paths explicitly defined (instead of automagically defined).

RomanHotsiy: One benefit is you can name your path files whatever you want not as long as `$ref` stays in-tact. You can also group them into folders if needed.

.gitignore
```diff
@@ -34,3 +34,7 @@ node_modules
 
 # npm
 package-lock.json
+
+
+# OpenAPI output bundles folder
+dist
```
ashkarpetin: no newline

openapi/openapi.yaml
slavcodev: What about definitions? Should they also be manually added?

RomanHotsiy: No, you can reference them directly from the places you use them. They will be pulled out to `components/schemas` section on bundle by filename.

E.g.: https://github.com/Rebilly/RebillyAPI/pull/796/files#diff-31e92a68a16eb684efa19ce85ffcd4ebR65

PR #800, opened 2020-01-21
Allow changing quantity in Change Plan

XaosSintez:


spec/components/schemas/SubscriptionChange.yaml
```diff
@@ -8,6 +8,9 @@ properties:
     description: The plan identifier string
     allOf:
       - $ref: "#/components/schemas/ResourceId"
+  quantity:
```
destebang: Not related with this RFC, but I noticed we use different behaviour for nullability in other parts of the API: https://github.com/Rebilly/RebillyAPI/pull/795/files#diff-11abc6233c34a6f9bd956593214383f7R103 
Personally, I prefer the behaviour in this PR and the other is more opinionated, but we should be consistent about how the API behaves in this situation

2e3s: Everything that is not required can be `null` (or omitted).
https://swagger.io/docs/specification/describing-parameters/
> Note: nullable is not the same as an optional parameter or an empty-valued parameter. nullable means the parameter value can be null. Specific implementations may choose to map an absent or empty-valued parameter to null, but strictly speaking these are not the same thing.

I read it like "null has a meaning".

PR #805, opened 2020-02-18
Add AML search API

mclassic:
Add API for AML searching


mattsanf (dismissed):
My understanding of OpenAPI spec is weak, so I'm not sure if any of the properties are missing any flags or if the ones added are unnecessarily. 

But the properties in the schema makes sense to me, from their type, name and description.

openapi/components/schemas/AML.yaml
```diff
@@ -0,0 +1,128 @@
...
+  lastName:
+    type: string
+    description: Last Name
+  source:
+    type: string
+    readOnly: true
+    description: Which list this came from.
+  sourceType:
+    readOnly: true
+    description: The list type
```
agalstian: may be that should be the field name? `listType`?

mclassic: If I rename it I'll have to rename `source` above it to `list` as well. What do others think?

openapi/paths/aml.yaml
```diff
@@ -0,0 +1,49 @@
+parameters:
+  - $ref: ../components/parameters/organizationId.yaml
+post:
```
mclassic: I originally made this a `POST` but semantically it should be a `GET`. Should I change it to `GET`?

openapi/paths/aml.yaml
```diff
@@ -0,0 +1,49 @@
...
+          $ref: ../components/headers/Pagination-Limit.yaml
+        Pagination-Offset:
+          $ref: ../components/headers/Pagination-Offset.yaml
+      content:
+        application/json:
+          schema:
+            type: array
+            items:
+              $ref: ../components/schemas/AML.yaml
+    '204':
```
mclassic: I like the idea of returning `204` instead of `404`. This makes sense to me, as the search was successful, it just didn't return hits.

openapi/components/schemas/AML.yaml
```diff
@@ -0,0 +1,128 @@
...
+    enum:
+      - pep
+      - sanctions
+      - adverse-media
+  gender:
+    type: string
+    readOnly: true
+    nullable: true
+    description: Gender of returned identity (if `type` is `individual`)
+  title:
```
adamaltman: Is this like "Mr."?

openapi/components/schemas/AML.yaml
```diff
@@ -0,0 +1,128 @@
+type: object
+properties:
```
adamaltman: Where would aliases go?

openapi/openapi.yaml
```diff
@@ -227,6 +233,7 @@ tags:
 x-tagGroups:
   - name: Customers
     tags:
+      - AML
```
adamaltman: Put this lower.  This renders in the order shown here.  I would put it under `KYC Documents`.

openapi/paths/aml.yaml
adamaltman: Yes, I wonder why it isn't a GET request?

Also, I wonder why a request body instead of query parameters?

openapi/paths/aml.yaml
adamaltman: I wonder why a 204 instead of a 200 like we do with all of our other GET collections APIs with an empty array if there are no hits?

openapi/paths/aml.yaml
mclassic: That's fair, i'll move it to GET and that will also include the move to query parameters. I'm not sure what I was thinking to be honest. :D

openapi/paths/aml.yaml
mclassic: I can do that. Not a problem.

openapi/components/schemas/AML.yaml
mclassic: "President", "General", "Colonel" etc. Most lists provide these types of things as titles, but not necessarily "Mr." or "Mrs."

openapi/components/schemas/AML.yaml
mclassic: I will add them as an array of objects. This was supposed to be in here. Thankful for reviews.

openapi/components/schemas/AML.yaml
adamaltman: Okay, so it may be like title of the position... `Mayor`? I think you should add an example, "President" is a clear enough example.

openapi/components/schemas/AML.yaml
```diff
@@ -0,0 +1,150 @@
...
+  firstName:
+    type: string
+    description: First Name
```
adamaltman: Examples may be good here too... such as Benjamin Franklin, etc..

openapi/components/schemas/AML.yaml
```diff
@@ -0,0 +1,150 @@
...
+  gender:
+    type: string
+    readOnly: true
+    nullable: true
+    description: Gender of returned identity (if `type` is `individual`)
+  title:
+    type: array
+    readOnly: true
+    nullable: true
+    description: The identity's title of their position, i.e. `President` or `General`
```
adamaltman:
```suggestion
    description: The identity's title of their position.
    example: President
```
I would make an example.  This is another line like so (if GitHub let's me add a line).

openapi/components/schemas/AML.yaml
adamaltman:
```suggestion
    description: First Name
    example: Benjamin
```

openapi/components/schemas/AML.yaml
```diff
@@ -0,0 +1,150 @@
+type: object
+properties:
+  firstName:
+    type: string
+    description: First Name
+  lastName:
+    type: string
+    description: Last Name
```
adamaltman:
```suggestion
    description: Last Name
    example: Franklin
```

openapi/components/schemas/AML.yaml
```diff
@@ -0,0 +1,150 @@
...
+    description: The nationality of the identity
+  address:
+    type: array
+    readOnly: true
+    description: Addresses related to the identity
+    items:
+      type: object
+      properties:
+        address:
+          type: string
```
adamaltman: these are all readOnly too, right?

mclassic: Yes

openapi/components/schemas/AML.yaml
```diff
@@ -0,0 +1,164 @@
...
+    enum:
+      - individual
+      - entity
+  legalBasis:
+    type: array
+    readOnly: true
+    description: List of references to legal documents if they exist
+    items:
+      type: string
+  regime:
```
ashkarpetin: Can you please give some examples?

mclassic: I added `ISIL` as an example just now.

openapi/components/schemas/AML.yaml
```diff
@@ -0,0 +1,165 @@
...
+    type: string
+    readOnly: true
+    description: Regime
+    nullable: true
+    example: ISIL
```
adamaltman: This isn't a good example. It will display Benjamin Franklin and ISIL on the example.  If we're using Benjamin Franklin throughout, he was part of the US.  

His political positions were Ambassador and Speaker of the House.

mclassic: @adamaltman But that would be under `title` if it was to be populated. Maybe Regime would be `US Government` but that sounds odd imo.

mclassic: Actually `US Government` would make sense.

openapi/components/schemas/AML.yaml
```diff
@@ -0,0 +1,165 @@
...
+    type: string
+    readOnly: true
+    nullable: true
+    description: Gender of returned identity (if `type` is `individual`)
+  title:
+    type: array
+    readOnly: true
+    nullable: true
+    description: The title of their position.
+    example: President
```
adamaltman: This is supposed to be an array of strings.
```suggestion
    example: 
      - Postmaster General
      - Ambassador to France
      - Ambassador to Sweden
      - Supreme Executive Council of Commonwealth of Pennsylvania
```

openapi/components/schemas/AML.yaml
```diff
@@ -0,0 +1,165 @@
...
+          description: This address is their birth place
+  dob:
+    type: array
+    readOnly: true
+    description: One or more possible dates of birth
+    items:
+      type: string
+      readOnly: true
+      description: Date of birth
+      format: date
```
adamaltman:
```suggestion
      format: date
      example: 1706-01-17
```

mclassic:
php please re-review
js please re-review

mclassic:
js please re-review

ashkarpetin (approved):

2e3s (approved):

openapi/components/schemas/AML.yaml
```diff
@@ -0,0 +1,170 @@
...
+    readOnly: true
+    enum:
+      - strong
+      - weak
+      - unknown
+  nationality:
+    type: string
+    readOnly: true
+    description: The nationality of the identity
+  address:
```
mattsanf:
```suggestion
  addresses:
```

mclassic:
No longer used repo.

PR #809, opened 2020-02-26
Invalid API description for Lead source API (GWY-14895)

vykhrystiuk:
## Details

LeadSource POST is incorrect and unused it should be renamed to PUT


openapi/paths/customers@{id}@lead-source.yaml
```diff
@@ -30,7 +30,7 @@ get:
     - lang: PHP
       source:
         $ref: '../code_samples/PHP/customers@{id}@lead-source/get.php'
-post:
+put:
```
VladimirBerdnik: Should https://github.com/Rebilly/RebillyAPI/pull/809/files#diff-768ada634c99e1fea3017b424f3bd745R37

and https://github.com/Rebilly/RebillyAPI/pull/809/files#diff-768ada634c99e1fea3017b424f3bd745R68

be updated as well?

VladimirBerdnik: Ok, I see, changes were just done

vykhrystiuk: I think so. Updated

vykhrystiuk:
php please review

PR #810, opened 2020-02-27
Add disputeIds to transaction resource

ashkarpetin:


openapi/components/schemas/Transactions/Transaction.yaml
```diff
@@ -402,6 +402,10 @@ properties:
     type: string
     description: The payment description
     maxLength: 255
+  disputeId:
```
ashkarpetin: There can be more than 1 dispute, so it seem it should be disputeIds

wmdmgv: so is it an array?

ashkarpetin: I've changed to disputeIds

PR #812, opened 2020-03-09
Add lineItemSubtotal to Subscription

XaosSintez:


openapi/components/schemas/Subscription.yaml
```diff
@@ -136,6 +136,13 @@ properties:
     nullable: true
     example: PO123456
     type: string
+  lineItemSubtotal:
+    description: >-
+      Subtotal of line items in this subscription (signed value).  If credits
```
edmandiesamonte:
```suggestion
      Subtotal of line items in this subscription (signed value). If credits
```

vykhrystiuk:
Could you provide a little bit more info about what this PR should do? `lineItemSubtotal` should be changed and it should be a money object that contains amount and currency, shouldn't it? If so, are there any backend changes?

openapi/components/schemas/Subscription/OrderTypes/subscription-order.yaml
```diff
@@ -111,6 +111,25 @@ allOf:
...
+        readOnly: true
+        description: >-
+          Subtotal of line items in this subscription (signed value). If credits
+          exceed debits, it will be a negative number.
+        properties:
+          currency:
+            allOf:
+              - $ref: ../../CurrencyCode.yaml
+          amount:
+            type: number
```
vykhrystiuk:
```suggestion
            type: number
            format: double
```
?

openapi/components/schemas/Subscription/OrderTypes/subscription-order.yaml
```diff
@@ -111,6 +111,26 @@ allOf:
...
```
vykhrystiuk: Do we need a description for amount and currency ?

XaosSintez: I don't think so. Do you think we need them?

XaosSintez:
@vykhrystiuk Backend already has such object in response. This PR should fix its definition. Previously it was incorrect, it had only amount, but without currency.

openapi/components/schemas/Subscription/OrderTypes/subscription-order.yaml
vykhrystiuk: Oh, I just noticed we have lineItemSubtotal description. I think no needs to do that.

openapi/components/schemas/Subscription/OrderTypes/subscription-order.yaml
```diff
@@ -111,6 +111,26 @@ allOf:
...
+          - $ref: ../UpcomingInvoiceItemCollection.yaml
+      lineItemSubtotal:
+        type: object
+        readOnly: true
+        description: >-
+          Subtotal of line items in this subscription (signed value). If credits
+          exceed debits, it will be a negative number.
+        properties:
+          currency:
+            allOf:
```
adamaltman: why the allOf?

XaosSintez: It contains `currency` definition. We use it everywhere:
https://github.com/Rebilly/RebillyAPI/blob/a9b4fbd62c8215df74af7437af90c9497e8fcccf/openapi/components/schemas/Payment.yaml#L21

https://github.com/Rebilly/RebillyAPI/blob/a9b4fbd62c8215df74af7437af90c9497e8fcccf/openapi/components/schemas/Invoices/Invoice.yaml#L36

https://github.com/Rebilly/RebillyAPI/blob/a9b4fbd62c8215df74af7437af90c9497e8fcccf/openapi/components/schemas/Transactions/Transaction.yaml#L53

slavcodev: @XaosSintez but `allOf` is used to extend the other definition, it's not the case here.
```
currency:
  $ref: ../../CurrencyCode.yaml
```
This works well

XaosSintez: Okay, fixed it. Thanks

